PR #17, opened 2020-04-02
Support httpapi in resource modules, part 1

Qalthos:
##### SUMMARY

This covers modifying the tests from nxos_interfaces to ???

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

httpapi/nxos


tests/integration/targets/nxos_interfaces/tasks/cli.yaml
```diff
@@ -1,21 +1,27 @@
 ---
 - name: collect common test cases
   find:
-    paths: '{{ role_path }}/tests/cli'
+    paths: '{{ role_path }}/tests/common'
     patterns: '{{ testcase }}.yaml'
   connection: local
   register: test_cases
 
+- name: collect cli test cases
```
NilashishC, line 9: All the tests have been moved under the common folder. Do we still need to keep this task?

tests/integration/targets/nxos_interfaces/tasks/nxapi.yaml
```diff
@@ -21,8 +21,7 @@
   set_fact: test_items="{{ test_cases.files | map(attribute='path') | list }}"
```
NilashishC, line 21: Maybe we can remove `collect nxapi test cases` task?

NilashishC:
Looks good to me. Just added two minor comments.

tests/integration/targets/nxos_interfaces/tasks/cli.yaml
Qalthos: This is consistent with the rest of the modules, and makes adding transport-specific tests easy if it becomes important later. I'm not saying it's necessary or even likely that we will have transport-specific tests, but the cost is practically nil for not having to fiddle with these files again if we do.

NilashishC: Yeah, makes sense.

PR #22, opened 2020-04-10
Remove starts with from error regex

NilashishC:

##### SUMMARY
- The response received doesn't start with (Error|error), instead in most cases it starts with the last command sent. As a result, the terminal error regex fails to catch it and the task doesn't fail. This PR removes the starts with from the regex.

```
response = b'redistribute bgp 65589 route-map rmap_8\r\r\nError: bgp-65578 is already distributed, Only one is allowed\r\r\nNexus9000v(config-router-vrf)#
```

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
terminal/nxos.py

### DIFF
--- a/plugins/terminal/nxos.py
+++ b/plugins/terminal/nxos.py
@@ -39,7 +39,7 @@ class TerminalModule(TerminalBase):
 
     terminal_stderr_re = [
         re.compile(br"% ?Error"),
-        re.compile(br"^error:(.*)", re.I),
+        re.compile(br"\nerror:(.*)", re.I),
         re.compile(br"^% \w+", re.M),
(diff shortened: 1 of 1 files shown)


NilashishC:
recheck

NilashishC:
recheck

NilashishC:
> This might result in regex match within the response on any `error: ` string even if it is not an error.
> Changing to `re.compile(br"\nerror:(.*)", re.I)` might be a more safer option.

Agreed. Pushed the changes.

NilashishC:
recheck

NilashishC:
recheck

NilashishC:
recheck

PR #31, opened 2020-04-22
Streamline tests and a small facts fix

NilashishC:

##### SUMMARY
- Some test changes + a facts fix.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
nxos_acl_interfaces

### DIFF
--- a/tests/integration/targets/nxos_acl_interfaces/tests/cli/populate_config.yaml
+++ b/tests/integration/targets/nxos_acl_interfaces/tests/cli/populate_config.yaml
@@ -1,7 +1,4 @@
 ---
 - name: Adding base configuration
-  ansible.netcommon.cli_config:
-    config: "ip access-list ACL1v4\nip access-list NewACLv4\nipv6 access-list ACL1v6\n\
-      ipv6 access-list NewACLv6\nip access-list PortACL\ninterface Ethernet1/2\n\
-      \  ipv6 traffic-filter ACL1v6 in\ninterface Ethernet1/5\n  ip port access-group\
-      \ PortACL in\n  ip access-group ACL1v4 out\n  ipv6 traffic-filter ACL1v6 in\n"
+  cisco.nxos.nxos_config:
+    src: populate_config.cfg
--- a/tests/integration/targets/nxos_acl_interfaces/tests/cli/remove_config.yaml
+++ b/tests/integration/targets/nxos_acl_interfaces/tests/cli/remove_config.yaml
@@ -1,12 +1,37 @@
 ---
-- name: Remove config
-  ansible.netcommon.cli_config:
-    config: "no ip access-list ACL1v4\nno ip access-list NewACLv4\nno ip access-list\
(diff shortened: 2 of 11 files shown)


tests/integration/targets/nxos_acl_interfaces/tests/cli/populate_config.yaml
```diff
@@ -1,7 +1,16 @@
 ---
 - name: Adding base configuration
-  ansible.netcommon.cli_config:
-    config: "ip access-list ACL1v4\nip access-list NewACLv4\nipv6 access-list ACL1v6\n\
-      ipv6 access-list NewACLv6\nip access-list PortACL\ninterface Ethernet1/2\n\
-      \  ipv6 traffic-filter ACL1v6 in\ninterface Ethernet1/5\n  ip port access-group\
-      \ PortACL in\n  ip access-group ACL1v4 out\n  ipv6 traffic-filter ACL1v6 in\n"
+  cisco.nxos.nxos_command:
+    commands:
+      - "configure"
```
Qalthos, line 5: I don't like this vs having two nxos_config plays, but if this is appreciably faster than that (especially over the number of times we run it) then I guess it's fine.

I would however add a comment describing the problem (and link to an issue tracking it?)

tests/integration/targets/nxos_acl_interfaces/tests/cli/remove_config.yaml
```diff
@@ -1,12 +1,8 @@
...
-      \ PortACL\nno ipv6 access-list ACL1v6\nno ipv6 access-list NewACLv6\ninterface\
-      \ Ethernet1/2\n  no ipv6 traffic-filter ACL1v6 in\ninterface Ethernet1/5\n\
-      \  no ip access-group ACL1v4 out\n  no ip port access-group PortACL in\n \
-      \ no ipv6 traffic-filter ACL1v6 in\ninterface Ethernet1/3\n  no ipv6 port\
-      \ traffic-filter NewACLv6 in\n  no ip access-group ACL1v4 out\n  no ip port\
-      \ access-group PortACL in\ninterface Ethernet1/4\n  no ip access-group NewACLv4\
-      \ out\n"
-  ignore_errors: True
+- name: Remove ACL interfaces
+  cisco.nxos.nxos_acl_interfaces:
```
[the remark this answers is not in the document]
NilashishC: You're right. I've updated it to use nxos_config.

tests/integration/targets/nxos_acl_interfaces/tests/cli/populate_config.yaml
NilashishC: Thinking out loud, the time taken by the extra tasks is probably very small in the grand scheme of things and using `src` for remove_config turned out to be risky in some cases leading to leftovers in config that cause assertions to fail. This is mostly because NX-OS throws an error if you try to remove an ACL that is not already configured on an interface and as a result, the subsequent negate lines are not sent to the appliance.

NilashishC:
recheck

Issue #27, opened 2020-04-20
nxos_user module returns changed even when command fails


garamirseokim:
##### SUMMARY

When the NX-OS device is configured with 'password strength-check' option and the nxos_user module attempts to create a new user and password on the device with a password that does not meet the strength requirements, Ansible reports the outcome as 'changed' when in fact the password was not created. When trying to change passwords, have the same issue as well.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

cisco.nxos.nxos_task

##### ANSIBLE VERSION

```paste below
Tested with below versions... 
Ansible 2.8.0,2.9.2 and 2.9.6
```

##### CONFIGURATION

```paste below

```

##### OS / ENVIRONMENT

```paste below
Tested with below versions... 
NXOS: version 7.0(3) and 9.3(3)
```
##### STEPS TO REPRODUCE

```
------------------------------------------------------
Preparation
------------------------------------------------------
test-n9k# show version | grep NXOS
 NXOS: version 9.3(3)
  NXOS image file is: bootflash:///nxos.9.3.3.bin
  NXOS compile time:  12/22/2019 2:00:00 [12/22/2019 14:00:37]
test-n9k# show password strength-check 
Password strength check is enabled
test-n9k# show user-account ansible
user:ansible does not exist
------------------------------------------------------
Playbook Execution
------------------------------------------------------
$ cat playbook/nxos_test.yml
---
- hosts: cisco_sandbox_nxos
  gather_facts: false
  tasks:
    - cisco.nxos.nxos_user:
        name: ansible
        state: present
        update_password: always
        configured_password: 'abc'
      register: status
    - debug:
        var: status
$ ansible --version
ansible 2.9.6
  config file = /home/foo/work/testing/ansible/ansible.cfg
  configured module search path = ['/home/foo/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.7/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.7.6 (default, Jan 30 2020, 09:44:41) [GCC 9.2.1 20190827 (Red Hat 9.2.1-1)] 

$ ansible-playbook -i inventory playbook/nxos_test.yml -v
Using /home/foo/work/testing/ansible/ansible.cfg as config file
PLAY [cisco_sandbox_nxos] ************************************************************************************************************************************
TASK [cisco.nxos.nxos_user] **********************************************************************************************************************************
[...]

PR #32, opened 2020-04-23
nxos_user should fail when device rejects configured_password

NilashishC:

##### SUMMARY
- Fixes https://github.com/ansible-collections/cisco.nxos/issues/27
- Remove superfluous return values from doc string

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
nxos_user.py

### DIFF
--- a/plugins/modules/nxos_user.py
+++ b/plugins/modules/nxos_user.py
@@ -118,21 +118,6 @@ commands:
   sample:
     - name ansible
     - name ansible password password
-start:
-  description: The time the job started
-  returned: always
-  type: str
-  sample: "2016-11-16 10:38:15.126146"
-end:
-  description: The time the job ended
-  returned: always
-  type: str
-  sample: "2016-11-16 10:38:25.595612"
-delta:
-  description: The time elapsed to perform all operations
-  returned: always
-  type: str
-  sample: "0:00:10.469466"
 """
 from copy import deepcopy
 from functools import partial
@@ -383,7 +368,7 @@ def main():
         supports_check_mode=True,
     )
 
-    result = {"changed": False}
+    result = {"changed": False, "warnings": []}
 
     want = map_params_to_obj(module)
     have = map_config_to_obj(module)
@@ -407,7 +392,19 @@ def main():
 
     if commands:
         if not module.check_mode:
-            load_config(module, commands)
+            responses = load_config(module, commands)
+            for resp in responses:
+                if resp.lower().startswith("wrong password"):
+                    module.fail_json(msg=resp)
+                else:
+                    result["warnings"].extend(
+                        [
+                            x[9:]
+                            for x in resp.splitlines()
+                            if x.startswith("WARNING: ")
+                        ]
+                    )
+
         result["changed"] = True
 
     module.exit_json(**result)
--- a/tests/integration/targets/nxos_user/tests/common/basic.yaml
+++ b/tests/integration/targets/nxos_user/tests/common/basic.yaml
@@ -2,7 +2,7 @@
 - debug: msg="START connection={{ ansible_connection }} nxos_user basic test"
 
 - name: Remove old entries of user
-  cisco.nxos.nxos_user:
+  cisco.nxos.nxos_user: &rem
     aggregate:
 
       - name: ansibletest1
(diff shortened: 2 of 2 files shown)


NilashishC:
recheck

tests/integration/targets/nxos_user/tests/common/basic.yaml
```diff
@@ -42,21 +44,23 @@
...
-
-      - name: ansibletest1
+    name: ansibletest_failed
+    configured_password: abc
+  register: result
+  ignore_errors: True
 
-      - name: ansibletest2
+- assert:
+    that: result.failed == True
```
trishnaguha, line 55: Will you be able to validate the error msg here?

NilashishC: Added that.

plugins/modules/nxos_user.py
```diff
@@ -407,7 +392,12 @@ def main():
 
     if commands:
         if not module.check_mode:
-            load_config(module, commands)
+            responses = load_config(module, commands)
+            for resp in responses:
+                if resp.lower().startswith("warning"):
+                    result["warnings"].append(resp)
```
Qalthos, line 398: This would presumably result in something like `WARNING: Warning something something` in the log right? Should we strip the "warning" prefix from response here?

For that matter, do you have something to trigger this branch? We should probably have a test that hits this.

NilashishC: Yeah, that's right. I thought the same, but didn't want to mess with the actual message received from the device. On second thoughts, this may not be a bad idea. I'll change this and add a test.

NilashishC:
@trishnaguha @Qalthos Seems like a single command issued may result in [multiple warnings](https://object-storage-ca-ymq-1.vexxhost.net/v1/a0b4156a37f9453eb4ec7db5422272df/ansible_32/32/5191dc1d51605c85c626826180e3ad809a2bb992/check/ansible-test-network-integration-nxos-cli-python36/74eb0c4/controller/ara-report/result/84ad036d-641b-4582-8cc4-3fccca6d6aba/).
So I made some changes to accommodate that and added tests. Warning responses always start with `WARNING:`, so the lower() seemed redundant. We can use a re.sub() to make this case insensitive but I found this behaviour to be consistent across 3 different NX-OS versions.

**Example:**

Nexus9000v(config)# username test password abc               
WARNING: Minimum recommended length of 8 characters.                 
WARNING: Password should contain characters from at least three of the following classes: lower case letters, upper case letters, digits and special characters.                                                 
WARNING: it is WAY too short
WARNING: Configuration accepted because password strength check is disabled

```
[WARNING]: Minimum recommended length of 8 characters.
[WARNING]: Password should contain characters from at least three of the following classes: lower case letters, upper case letters, digits and special characters.
[WARNING]: it is WAY too short
[WARNING]: Configuration accepted because password strength check is disabled
```

NilashishC:
CI is green for nxos_user - https://object-storage-ca-ymq-1.vexxhost.net/v1/a0b4156a37f9453eb4ec7db5422272df/ansible_32/32/6a8321ee34a70423b246e0e4023771a1aad3209d/check/ansible-test-network-integration-nxos-cli-python36/67dd1bf/controller/ara-report/reports/b2b5423b-4084-4e44-b59a-880452542201.html

plugins/modules/nxos_user.py
```diff
@@ -407,7 +392,14 @@ def main():
 
     if commands:
         if not module.check_mode:
-            load_config(module, commands)
+            responses = load_config(module, commands)
+            for resp in responses:
+                if resp.startswith("WARNING: "):
+                    result["warnings"].extend(
+                        [x.replace("WARNING: ", "") for x in resp.splitlines()]
+                    )
```
Qalthos, line 400: I don't know if warnings can come after other response lines, but this seems more robust in general.

```suggestion
                result["warnings"].extend([
                    x[9:] for x in resp.splitlines()
                    if x.startswith("WARNING: ")
                ])
```

NilashishC: I hadn't thought of this. This would probably not happen, but in any case what you suggested looks much better. I'll push these changes.

NilashishC:
recheck

PR #33, opened 2020-04-23
Add nxos_ospfv2 Resource Module

NilashishC:

##### SUMMARY
- Add nxos_ospfv2 resource module.
- Base classes will most definitely be moved to a common place.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
nxos_ospfv2.py


NilashishC:
CI green for nxos_ospfv2 - https://object-storage-ca-ymq-1.vexxhost.net/v1/a0b4156a37f9453eb4ec7db5422272df/ansible_33/33/409ce31a78ead5120ca76adebd44cca94fabf2bb/check/ansible-test-network-integration-nxos-cli-python36/821f433/controller/ara-report/reports/53ebbe02-dc18-4a9e-b6f2-435504436372.html

.gitignore
```diff
@@ -102,3 +102,7 @@ venv.bak/
 
 # mypy
 .mypy_cache/
+
+# ide
```
justjais, line 106: is it required?

plugins/module_utils/network/common/network_template.py
```diff
@@ -0,0 +1,94 @@
+import re
+from copy import deepcopy
+
+from ansible_collections.ansible.netcommon.plugins.module_utils.network.common.utils import (
+    Template,
+    dict_merge,
+)
+
+
+class NetworkTemplate(object):
```
justjais, line 10: can we move the netcommon changes to diff PR, as it can be referred separately in future

plugins/module_utils/network/nxos/config/ospfv2/ospfv2.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def __init__(self, module):
+        super(Ospfv2, self).__init__(
+            empty_fact_val={},
+            facts_module=Facts(module),
+            module=module,
+            resource="ospfv2",
+            tmplt=Ospfv2Template(),
+        )
+        self.parsers = [
```
justjais, line 44: is it required to initialize parsers value in __init__ as not being used by any other fns except _compare, ideally it should be declared under _compare fn. itself

plugins/module_utils/network/nxos/rm_templates/ospfv2.py
```diff
@@ -0,0 +1,1099 @@
...
+    command = "area {area_id} nssa translate type7".format(**area)
+    for attrib in ["always", "never", "supress_fa"]:
+        if translate.get(attrib):
+            command += " {0}".format(attrib.replace("_", "-"))
+    return command
+
+
+def _tmplt_area_ranges(arange):
+    command = "area {area_id} range {prefix}".format(**arange)
+    if arange.get("not_advertise") is True:
```
justjais, line 147: is it required to explicitly check for True

plugins/module_utils/network/nxos/rm_templates/ospfv2.py
```diff
@@ -0,0 +1,1099 @@
...
+        command += " not-advertise"
+    if "cost" in arange:
+        command += " cost {cost}".format(**arange)
+    return command
+
+
+def _tmplt_area_ranges(arange):
+    command = "area {area_id} range {prefix}".format(**arange)
+    if arange.get("not_advertise") is True:
```
justjais, line 156: is it required to explicitly check for True

plugins/module_utils/network/nxos/rm_templates/ospfv2.py
```diff
@@ -0,0 +1,1099 @@
...
+        command += " tag {tag}".format(**proc)
+    elif proc.get("not_advertise"):
+        command += " not-advertise"
+    return command
+
+
+def _tmplt_area_stub(area):
+    stub = area["stub"]
+    command = "area {area_id} stub".format(**area)
+    if stub.get("set") is False:
```
justjais, line 175: is it required to explicitly check for False
```suggestion
    if not stub.get("set"):
```

plugins/module_utils/network/nxos/rm_templates/ospfv2.py
```diff
@@ -0,0 +1,1099 @@
...
+        command += " {id}".format(**redis)
+    if redis.get("route_map"):
+        command += " route-map {route_map}".format(**redis)
+    return command
+
+
+def _tmplt_capability_vrf_lite(proc):
+    command = "capability vrf-lite"
+    vrf_lite = proc["capability"]["vrf_lite"]
+    if vrf_lite.get("set") is False:
```
justjais, line 194: is it required to explicitly check for False

plugins/modules/nxos_ospfv2.py
```diff
@@ -0,0 +1,1610 @@
...
+                    no_summary: True
+                    translate:
+                      type7:
+                        always: True
+            - vrf: zone2
+              auto_cost:
+                reference_bandwidth: 45
+                unit: Gbps
+    state: merged
+
```
justjais, line 1072: can u also include command fired, as it gives more visibility to the users for all of the valid states

tests/integration/targets/nxos_ospfv2/meta/main.yml
```diff
@@ -0,0 +1 @@
+
```
justjais, line 1: space can be removed

tests/integration/targets/nxos_ospfv2/tests/common/_populate_config.yaml
```diff
@@ -0,0 +1,26 @@
+- name: Setup
+  cisco.nxos.nxos_config:
+    lines:
```
justjais, line 3: IMO it's better to use overridden state for setting configuration instead of line

tests/integration/targets/nxos_ospfv2/vars/main.yml
```diff
@@ -0,0 +1,158 @@
...
+      - process_id: '104'
+        router_id: 203.0.113.20
+      - process_id: '102'
+        router_id: 198.51.100.1
+        shutdown: True
+deleted:
+  after:
+    processes:
+      - process_id: '100'
+        router_id: 203.0.113.20
```
justjais, line 158: Unit testing seems to be missing

plugins/module_utils/network/nxos/config/ospfv2/ospfv2.py
```diff
@@ -0,0 +1,232 @@
...
+        for entry in wantd, haved:
+            self._ospf_list_to_dict(entry)
+
+        # if state is merged, merge want onto have and then compare
+        if self.state == "merged":
+            wantd = dict_merge(haved, wantd)
+
+        # if state is deleted, empty out wantd and set haved to wantd
+        if self.state == "deleted":
+            haved = {k: v for k, v in haved.items() if k in wantd or not wantd}
```
justjais, line 111: also, we need to change all of the `.items()` to ansible `iteritems()` instead
```suggestion
            haved = {k: v for k, v in iteritems(haved) if k in wantd or not wantd}
```

tests/integration/targets/nxos_ospfv2/tests/common/_populate_config.yaml
NilashishC: We shouldn't be using the same module for setup as the one we are testing.

plugins/module_utils/network/common/resource_module.py
```diff
@@ -0,0 +1,158 @@
+from copy import deepcopy
+from ansible.module_utils.connection import Connection
```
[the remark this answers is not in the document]
NilashishC: Yeah, good catch. These files will anyway be removed.

plugins/module_utils/network/nxos/config/ospfv2/ospfv2.py
NilashishC: Ah, right.

tests/integration/targets/nxos_ospfv2/vars/main.yml
NilashishC: Yeah, I haven't pushed those yet.

.gitignore
NilashishC: Yep. This is certainly useful for people working on VSCode.

plugins/module_utils/network/common/network_template.py
NilashishC: You are right. The netcommon PR (https://github.com/ansible-collections/ansible.netcommon/pull/21) is blocked by a NETCONF issue now.

plugins/module_utils/network/nxos/config/ospfv2/ospfv2.py
NilashishC: I find this approach to be much more cleaner than having a list as big as this within a function. 

Also, the original intention was to reuse a sub-list of the parser list for VRF comparison since not all the global keys are valid for VRF context. That way, we can decrease the run time a little bit by not making unnecessary comparison in base class compare().

NilashishC: I am yet to make that change. I'll push that along with the unit tests.

plugins/module_utils/network/nxos/rm_templates/ospfv2.py
NilashishC: Yes. Consider a scenario where "capability vrf-lite evpn" is not configured on the box and we run a task that looks has:
```yaml
- process_id: 100
   vrfs:
    - vrf: blue
      capability:
        vrf_lite:
          evpn: True
```

This should push "capability vrf-lite evpn", however, if the explicit check for False is in not given, it would instead send "no capability vrf-lite". This is because the dictionary that reaches this method will be something like `'capability': {'vrf_lite': {'evpn': True}}`, i.e., missing the "set" key and the method would fail to differentiate between absence of "set"
"set: False".

In any case, I feel having these explicit checks make the code easier to read and maintain.

Valid for all the other review comments regarding the same.

plugins/modules/nxos_ospfv2.py
NilashishC: Ack.

Issue #19, opened 2020-04-07
Support httpapi in reosurce modules


Qalthos:
##### SUMMARY

- [x] Add support to httpapi plugin (#17)
- [x] Migrate tests to cover both connections:
  - [x] nxos_acls (#38)
  - [x] nxos_acl_interfaces (#38)
  - [x] nxos_bfd_interfaces (#38)
  - [x] nxos_interfaces (#17)
  - [x] nxos_l2_interfaces (#24)
  - [x] nxos_l3_interfaces (#24)
  - [x] nxos_lacp (#24)
  - [x] nxos_lacp_interfaces (#24)
  - [x] nxos_lag_interfaces (#38)
  - [x] nxos_lldp_global (#24)
  - [x] nxos_lldp_interfaces (#24)
  - [x] nxos_telemetry (#38)
  - [x] nxos_vlans (#38)

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

httpapi/nxos.py

NilashishC:
@Qalthos - We also need to add nxos_bfd_global and nxos_telemetry to this list. Thanks!

Qalthos:
@NilashishC Did you mean `nxos_bfd_interfaces`? `nxos_bfd_global` already has nxapi support.

NilashishC:
> @NilashishC Did you mean `nxos_bfd_interfaces`? `nxos_bfd_global` already has nxapi support.

@Qalthos ah, right - bfd_interfaces.

Qalthos:
Done!

PR #38, opened 2020-04-27
 Support httpapi in resource modules, part 3

Qalthos:
##### SUMMARY

Modifies the tests for several modules

See also #19

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

nxos_acls
nxos_acl_interfaces
nxos_bfd_interfaces
nxos_lag_interfaces
nxos_telemetry
nxos_vlans


tests/integration/targets/nxos_bfd_interfaces/tests/common/overridden.yaml
```diff
@@ -13,16 +13,20 @@
   when: platform is not search('N5K|N6K')
 
 - name: setup1
-  ansible.netcommon.cli_config: &id002
+  cisco.nxos.nxos_config: &id002
     config: "no feature bfd\ndefault interface {{ test_int1 }}\ndefault interface\
```
NilashishC, line 17: In some of the changes, including this, the `config` key is not replaced with `lines`. This doesn't throw an error because in nxos_config module, the `config` key is an alias for `running_config`. As a result, these tasks actually don't push changes to the device.

Qalthos: Ah, good catch. I'll recheck and see what all I missed.

Qalthos:
recheck

Qalthos:
recheck

Qalthos:
recheck

Issue #36, opened 2020-04-27
Add new states to older Resource Modules


NilashishC:
##### SUMMARY
Add `gathered`, `rendered`, `parsed` support to the following RMs:
  - [x] nxos_hsrp_interfaces (#45)
  - [x] nxos_bfd_interfaces (#44)
  - [x] nxos_interfaces (#57)
  - [x] nxos_l2_interfaces (#37)
  - [x] nxos_l3_interfaces (#47)
  - [x] nxos_lacp (#39)
  - [x] nxos_lacp_interfaces (#41)
  - [x] nxos_lag_interfaces (#40)
  - [x] nxos_lldp_global (#42)
  - [x] nxos_vlans

##### ISSUE TYPE
- Feature Idea

trishnaguha:
Please take care of adding fully qualified collection name in the examples for these modules in the same PRs as well.

NilashishC:
Closing this since all the PRs are merged.

PR #44, opened 2020-05-04
Add new states to nxos_bfd_interfaces

NilashishC:

##### SUMMARY
Related to #36 

This also fixes an issue in config code. Since gather_subset was hardcoded to be ['!all', '!min], the platform version returned would always be empty. This patch changes it to ["min"] by default and for non-action states + gathered, it'd be ['!all', '!min'].

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
nxos_bfd_interfaces.py

### DIFF
--- a/plugins/modules/nxos_bfd_interfaces.py
+++ b/plugins/modules/nxos_bfd_interfaces.py
@@ -37,12 +37,24 @@ ANSIBLE_METADATA = {
 }
 
 DOCUMENTATION = """module: nxos_bfd_interfaces
-short_description: Manages BFD attributes of nxos interfaces.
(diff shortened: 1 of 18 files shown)


plugins/modules/nxos_bfd_interfaces.py
```diff
@@ -165,8 +277,18 @@ def main():
 
     :returns: the result form module invocation
     """
+    required_if = [
+        ("state", "merged", ("config",)),
+        ("state", "replaced", ("config",)),
+        ("state", "rendered", ("config",)),
```
rohitthakur2590, line 283: ("state", "overridden", ("config",)),

PR #49, opened 2020-05-05
Updated nxos.py terminal plugin on_become

MrMikeandike:
##### SUMMARY

Updated on_become to consistentsly strip whitespace before checking to see if prompt ends with "enable#".
Before, it ran 1 check without strip and 1 check with strip.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

plugins/terminal/nxos.py


NilashishC:
recheck

PR #58, opened 2020-05-12
httpapi enable issue

Qalthos:
##### SUMMARY

With delayed connection start, `self._become` _can_ change from the before the request to after. Rather than blindly stripping responses, from the output, decline to include responses whose input was enable.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

plugins/httpapi/nxos

### DIFF
--- a/plugins/httpapi/nxos.py
+++ b/plugins/httpapi/nxos.py
@@ -117,7 +117,9 @@ class HttpApi(HttpApiBase):
         device_info = {}
 
         device_info["network_os"] = "nxos"
-        reply, platform_reply = self.send_request(["show version", "show inventory"])
+        reply, platform_reply = self.send_request(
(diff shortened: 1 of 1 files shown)


NilashishC:
@Qalthos I don't know from where a "loopback1" interface sometimes leaks into the tasks.

https://object-storage-ca-ymq-1.vexxhost.net/v1/a0b4156a37f9453eb4ec7db5422272df/ansible_58/58/c67717f89fe2b57c29446e753b3d766574ceab8b/gate/ansible-test-network-integration-nxos-cli-python36_2_of_4/add61ca/controller/ara-report/reports/8e176cb6-3dd1-43db-bd57-31167f10b966.html

Can you explicitly clean up that up in remove config?

PR #76, opened 2020-06-01
Add nxos_smoke tests back

NilashishC:

##### SUMMARY
- Add nxos_smoke

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
targets/nxos_smoke

### DIFF
--- a/tests/integration/targets/nxos_smoke/tasks/main.yaml
+++ b/tests/integration/targets/nxos_smoke/tasks/main.yaml
@@ -1,0 +1,14 @@
+---
+# Some of the tests in this suite change the hostname to switch.
+# This block/always ensures the hostname gets changed back to
+# the correct name.
+- block:
+  - { include: cli.yaml, tags: ['cli'] }
+  - { include: nxapi.yaml, tags: ['nxapi'] }
+
+  always:
+  - name: "Change hostname back to {{ inventory_hostname_short }}"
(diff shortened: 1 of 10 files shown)


tests/integration/targets/nxos_smoke/tasks/main.yaml
```diff
@@ -0,0 +1,14 @@
+---
+# Some of the tests in this suite change the hostname to switch.
+# This block/always ensures the hostname gets changed back to
+# the correct name.
+- block:
+  - { include: cli.yaml, tags: ['cli'] }
+  - { include: nxapi.yaml, tags: ['nxapi'] }
```
Qalthos, line 7:
```suggestion
  - { include: cli.yaml, tags: ['network_cli'] }
  - { include: nxapi.yaml, tags: ['httpapi'] }
```

NilashishC: We have been using `cli` and `nxapi` tags in all the tests and in fact, CI uses `nxapi` tag to skip httapi/nxapi tests.

Qalthos: Alright, I had assumed that change was going to be applied to both, but I suppose this isn't the PR to enforce that. Carry on.

PR #91, opened 2020-06-13
Fixed #90, 8.4(2) introduced single-session feature, so pattern to ge…

### LINKED ISSUE #90: Unable to create Zones | NXOS 8.4(2)
##### SUMMARY
With switch running 8.4.2, zones are not getting created with zone module

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
nxos_zone_zoneset

##### STEPS TO REPRODUCE
Load switch with 8.4.2
Run nxos_zone_zoneset module to create zone, it will say zone lock acquired

##### EXPECTED RESULTS
zone should get created

##### ACTUAL RESULTS
module will say zone lock acquired

srbharadwaj:
Bug fix for #90

### DIFF
--- a/plugins/modules/storage/nxos_zone_zoneset.py
+++ b/plugins/modules/storage/nxos_zone_zoneset.py
@@ -916,18 +916,27 @@ def main():
                             + " is not activated, hence cannot deactivate"
                         )
                 elif actionflag == "activate":
-                    messages.append(
-                        "activating zoneset '"
-                        + zsetname
-                        + "' in vsan "
-                        + str(vsan)
-                    )
-                    actcmd.append(
-                        "zoneset activate name "
-                        + zsetname
-                        + " vsan "
-                        + str(vsan)
-                    )
+                    if commands_executed:
+                        messages.append(
+                            "activating zoneset '"
+                            + zsetname
+                            + "' in vsan "
+                            + str(vsan)
+                        )
+                        actcmd.append(
+                            "zoneset activate name "
+                            + zsetname
+                            + " vsan "
+                            + str(vsan)
+                        )
+                    else:
+                        messages.append(
+                            "no changes to existing zoneset '"
+                            + zsetname
+                            + "' in vsan "
+                            + str(vsan)
+                            + " hence activate action is ignored"
+                        )
             commands_executed = commands_executed + dactcmd + actcmd
 
         if commands_executed:
--- a/plugins/modules/storage/nxos_zone_zoneset.py
+++ b/plugins/modules/storage/nxos_zone_zoneset.py
@@ -362,7 +362,7 @@ class ShowZoneStatus(object):
 
         patfordefzone = "VSAN: " + str(self.vsan) + r" default-zone:\s+(\S+).*"
         patformode = r".*mode:\s+(\S+).*"
-        patforsession = r".*session:\s+(\S+).*"
+        patforsession = r"^session:\s+(\S+).*"
         patforsz = r".*smart-zoning:\s+(\S+).*"
         for line in output:
             if "is not configured" in line:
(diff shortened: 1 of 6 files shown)


tests/integration/targets/nxos_zone_zoneset/tests/common/sanity.yaml
```diff
@@ -144,13 +144,19 @@
...
       cisco.nxos.nxos_zone_zoneset: *id001
 
+    - name: Disaplay Idempotence Check result
+      debug:
+        var: result
+
     - assert:
         that:
-          - result.commands == []
+          - result.commands == ["terminal dont-ask", "zoneset activate name zsetname1
```
NilashishC, line 153: Why is an idempotent task sending commands?

NilashishC, line 153: Why is an idempotent task sending commands?

NilashishC:
recheck

tests/integration/targets/nxos_zone_zoneset/tests/common/sanity.yaml
srbharadwaj: zoneset activation is a bit unique it needs to be sent always

NilashishC:
recheck

tests/integration/targets/nxos_zone_zoneset/tests/common/sanity.yaml
```diff
@@ -144,13 +144,19 @@
...
+            vsan 922","zone commit vsan 922", "zoneset activate name zsetname21
+            vsan 923", "no terminal dont-ask"]
 
     - assert:
         that:
-          - result.changed == false
+          - result.changed == true
```
NilashishC, line 159: @srbharadwaj this task seems to have been idempotent previously. If the changes in this patch inevitably causes the module to loose idempotency in this particular use case, please add a note in the module doc stating the reason for future reference.

srbharadwaj: how to put it in module doc? any example?

srbharadwaj: @NilashishC pls dont merge this yet, we can make this idempotent. i'am working it, will test it out locally and then commit, till then pls keep this PR open

plugins/modules/storage/nxos_zone_zoneset.py
```diff
@@ -916,18 +916,27 @@ def main():
...
-                        + str(vsan)
-                    )
-                    actcmd.append(
-                        "zoneset activate name "
-                        + zsetname
-                        + " vsan "
-                        + str(vsan)
-                    )
+                    if commands_executed:
+                        messages.append(
```
NilashishC, line 920: Looks like this module also returns a key `messages` in the result. This has not been documented. Could you please add this [here](https://github.com/ansible-collections/cisco.nxos/blob/70b5242bc27961248df9a702ae712109f8911c05/plugins/modules/storage/nxos_zone_zoneset.py#L187)?

srbharadwaj: sure i also noticed that the mds modules were kept in plugins/modules/storage/ but i see copy of it in plugins/modules/ as well, should i delete those?

NilashishC: @srbharadwaj the ones in plugins/modules/ are just soft links to the ones in plugins/modules/storage.

NilashishC: They don't seem to be picked up by ansible from within the storage folder directly, so we had to add those.

Issue #110, opened 2020-07-03
nxos_lag_interfaces fails when force parameter is set 


Marcusz97:
##### SUMMARY
The nxos_lag_interfaces fails when the force parameter is set to either yes or no.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
nxos_lag_interfaces

##### ANSIBLE VERSION

```
ansible 2.9.10
  config file = /etc/ansible/ansible.cfg
  configured module search path = [u'/root/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.7/dist-packages/ansible
  executable location = /usr/bin/ansible
  python version = 2.7.17 (default, Apr 15 2020, 17:20:14) [GCC 7.5.0]
```

##### CONFIGURATION

```
DEFAULT_ASK_PASS(/Users/rchambers/dcnf/ansible.cfg) = True
DEFAULT_FORKS(/Users/rchambers/dcnf/ansible.cfg) = 10
DEFAULT_GATHERING(/Users/rchambers/dcnf/ansible.cfg) = explicit
DEFAULT_HOST_LIST(/Users/rchambers/dcnf/ansible.cfg) = ['/Users/rchambers/dcnf/fabric-lab.ini']
DEFAULT_MODULE_NAME(/Users/rchambers/dcnf/ansible.cfg) = command
HOST_KEY_CHECKING(/Users/rchambers/dcnf/ansible.cfg) = False
INTERPRETER_PYTHON(/Users/rchambers/dcnf/ansible.cfg) = auto_silent
PERSISTENT_COMMAND_TIMEOUT(/Users/rchambers/dcnf/ansible.cfg) = 30
PERSISTENT_CONNECT_RETRY_TIMEOUT(/Users/rchambers/dcnf/ansible.cfg) = 15
PERSISTENT_CONNECT_TIMEOUT(/Users/rchambers/dcnf/ansible.cfg) = 60
RETRY_FILES_ENABLED(/Users/rchambers/dcnf/ansible.cfg) = False

```

##### OS / ENVIRONMENT

Target host: Cisco Nexus NX-OSv, NXOS 9.3.3

##### STEPS TO REPRODUCE

```yaml
- name: add interface to the Port-channel
  nxos_lag_interfaces:
    config:
      - name: port-channel1
        members:
          - member: Ethernet1/4
            force: "true"
            mode: active
```

##### EXPECTED RESULTS

[...]

PR #111, opened 2020-07-07
Update lag_interfaces.py

Marcusz97:
fix error appending boolean to string when force keyword is hitten

##### SUMMARY

Fix issue #110 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

nxos_lag_interfaces

##### ADDITIONAL INFORMATION

When used force keyword the code tried to append the boolean value of the "force" option in the string to be send to the device, instead the expected behavior is that, if force is true, only prepare che string "channel-group X force [mode X]"

### DIFF
--- a/plugins/module_utils/network/nxos/config/lag_interfaces/lag_interfaces.py
+++ b/plugins/module_utils/network/nxos/config/lag_interfaces/lag_interfaces.py
@@ -286,8 +286,8 @@ class Lag_interfaces(ConfigBase):
             commands.append("interface" + " " + d["member"])
             cmd = ""
             group_cmd = "channel-group {0}".format(name)
-            if "force" in d:
-                cmd = group_cmd + " force " + d["force"]
+            if "force" in d and d["force"]:
+                cmd = group_cmd + " force "
             if "mode" in d:
                 if cmd:
                     cmd = cmd + " mode " + d["mode"]
--- a/plugins/module_utils/network/nxos/config/lag_interfaces/lag_interfaces.py
+++ b/plugins/module_utils/network/nxos/config/lag_interfaces/lag_interfaces.py
@@ -286,7 +286,7 @@ class Lag_interfaces(ConfigBase):
             commands.append("interface" + " " + d["member"])
             cmd = ""
             group_cmd = "channel-group {0}".format(name)
-            if "force" in d and d["force"]:
+            if d.get("force"):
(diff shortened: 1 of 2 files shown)


plugins/module_utils/network/nxos/config/lag_interfaces/lag_interfaces.py
```diff
@@ -286,8 +286,8 @@ def add_commands(self, diff, name):
             commands.append("interface" + " " + d["member"])
             cmd = ""
             group_cmd = "channel-group {0}".format(name)
-            if "force" in d:
-                cmd = group_cmd + " force " + d["force"]
+            if "force" in d and d["force"]:
```
Qalthos, line 289: Is this not sufficient?
```suggestion
            if d.get("force"):
```

Marcusz97: > Is this not sufficient?

Yes, but I'm worried about the existence of the keyword "force" in the dictionary.

Maybe is better? :

`if d.get("force",False):`

Qalthos: > Yes, but I'm worried about the existence of the keyword "force" in the dictionary.

That's what `.get()` is for? The default action is to return `None` when the key doesn't exist, which is falsy and so fails the conditional. Unless you're explicitly testing for `False`, which is unnecessary here, there's no reason to set the default value to anything else.

Marcusz97: Thank you, sorry I'm pretty new to this world.

NilashishC:
@Marcusz97 thanks a lot for fixing this! Can you please add a changelog to this PR? You can refer to [this](https://github.com/ansible/ansible/pull/70578/files#diff-b2db3cd21f72c2707ef77e0e74ba854bR1).

Marcusz97:
> @Marcusz97 thanks a lot for fixing this! Can you please add a changelog to this PR? You can refer to [this](https://github.com/ansible/ansible/pull/70578/files#diff-b2db3cd21f72c2707ef77e0e74ba854bR1).

@NilashishC Thank you! Is it ok like the one added?

NilashishC:
@Marcusz97 Looks good to me. Thanks!

PR #121, opened 2020-07-13
Make config modules work properly when module alias is used in task

NilashishC:

##### SUMMARY
- Module name will not always be prefixed with `{{ network_os }}_`.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
action/nxos.py


NilashishC:
> A more strict check can be added eg.
> `module_name in ['nxos_config', 'config']`
> 
> Please add integration/unit test and also, this requires to be fixed in other collections as well.

Fixed.

NilashishC:
recheck

NilashishC:
recheck

PR #125, opened 2020-07-21
Allow --diff and --check to coexist on nxos_config

Qalthos:
##### SUMMARY

Specifying both --check and --diff produces a warning and not a diff. This seems undesirable.
See also: ansible-collections/arista.eos#91, ansible-collections/cisco.ios#113

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

nxos_config


NilashishC:
recheck

PR #135, opened 2020-08-11
Add support for single_user_mode

NilashishC:

##### SUMMARY
- Targets https://github.com/ansible-collections/ansible.netcommon/issues/73
- Add nxos cliconf specific bits for caching feature

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/cliconf/nxos.py

### DIFF
--- a/plugins/cliconf/nxos.py
+++ b/plugins/cliconf/nxos.py
@@ -281,7 +281,6 @@ class Cliconf(CliconfBase):
         )
 
     def run_commands(self, commands=None, check_rc=True):
-        in_config_mode = False
         if commands is None:
             raise ValueError("'commands' value is required")
 
@@ -302,16 +301,10 @@ class Cliconf(CliconfBase):
             try:
                 out = self.send_command(**cmd)
                 if self.get_cli_prompt_context().endswith(")#"):
-                    if in_config_mode:
-                        # the command has been sent in configuration mode
-                        # if there is an existing cache, empty it
-                        if self._connection.get_cache():
-                            self._connection.get_cache().invalidate()
-                    else:
-                        # the device has entered configuration mode
-                        in_config_mode = True
-                else:
-                    in_config_mode = False
+                    # the device has entered configuration mode
+                    # if there is an existing cache, empty it
+                    if self._connection.get_cache():
+                        self._connection.get_cache().invalidate()
             except AnsibleConnectionFailure as e:
                 if check_rc is True:
                     raise
--- a/plugins/cliconf/nxos.py
+++ b/plugins/cliconf/nxos.py
@@ -45,7 +45,6 @@ from ansible_collections.ansible.netcommon.plugins.module_utils.network.common.u
     to_list,
 )
 from ansible.plugins.cliconf import CliconfBase, enable_mode
-from ansible.plugins.loader import cache_loader
 from ansible_collections.ansible.netcommon.plugins.cache.network import (
     NetworkCache,
 )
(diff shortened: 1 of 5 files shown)


pabelanger:
recheck

NilashishC:
recheck

NilashishC:
recheck

GomathiselviS:
recheck

NilashishC:
recheck

NilashishC:
Seems like the ansible/ansible depends-on PR is not being used?

NilashishC:
recheck

NilashishC:
recheck

pabelanger:
recheck

plugins/cliconf/nxos.py
```diff
@@ -213,7 +214,7 @@ def get_config(self, source="running", format="text", flags=None):
             cmd += " ".join(to_list(flags))
         cmd = cmd.strip()
 
-        return self.send_command(cmd)
+        return self.send_command(cmd, use_cache=True)
```
pabelanger, line 217: So, this is going to be an API change with ansible-base, which means we'll need to update the minimal version of ansible we are going to support.

Additionally, will this feature be backported into 2.10 and 2.9? If now, we likely need to make this feature flag condition in our collection.

pabelanger:
> Seems like the ansible/ansible depends-on PR is not being used?

I've fixed it, there was a bug in our ansible-zuul-jobs repo for ansible-test

NilashishC:
recheck

NilashishC:
recheck

NilashishC:
recheck

plugins/cliconf/nxos.py
ganeshrn: In order to reduce dependency on ansible-base we can copy the send_command API in network_cli connection plugin which is what the PR https://github.com/ansible-collections/ansible.netcommon/pull/120/files is doing so, in that case, we won't require backport or minimal ansible version dependency to be updated.

pabelanger: ++ I think that is the right approach, I guess we'll have to remove the bits in ansible/ansible right?

ganeshrn: Yes, it can be removed, but need to be careful doing so in order to ensure backward compatibility is maintained.

NilashishC:
recheck

plugins/cliconf/nxos.py
```diff
@@ -28,6 +28,15 @@
 - This nxos plugin provides low level abstraction apis for sending and receiving CLI
   commands from Cisco NX-OS network devices.
 version_added: 1.0.0
+options:
+  config_commands:
+    description:
+    - Specifies a list of commands that can make configuration changes
```
ganeshrn, line 34: The description can be more verbose as to why, where and how it is used.

plugins/cliconf/nxos.py
```diff
@@ -28,6 +28,15 @@
 - This nxos plugin provides low level abstraction apis for sending and receiving CLI
   commands from Cisco NX-OS network devices.
 version_added: 1.0.0
+options:
+  config_commands:
```
ganeshrn, line 32: also, add version_added string

NilashishC:
recheck

NilashishC:
recheck

NilashishC:
recheck

NilashishC:
Scheduled to be merged and released alongside ansible.netcommon v2.0.0.

NilashishC:
recheck

NilashishC:
recheck

NilashishC:
recheck

NilashishC:
recheck

NilashishC:
recheck

ashwini-mhatre:
LGTM

PR #139, opened 2020-08-18
N9K multisite feature

praveenramoorthy:
##### SUMMARY
Adding NXOS multisite support.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
nxos_bgp_af
nxos_bgp_neighbor
nxos_bgp_neighbor_af
nxos_vxlan_vtep
nxos_vxlan_vtep_vni
nxos_l3_interfaces


NilashishC (changes requested):

plugins/modules/nxos_bgp_af.py
```diff
@@ -213,6 +213,14 @@
     - present
     - absent
     type: str
+  retain_route_target:
```
NilashishC, line 216: Please add a `version_added: 1.1.0` key to all the new attributes.

NilashishC (changes requested):

plugins/modules/nxos_vxlan_vtep_vni.py
```diff
@@ -84,21 +84,30 @@
     choices:
     - present
     - absent
+    type: list
```
NilashishC, line 87: seems like the type of `state` key was changed to list?
```suggestion
    type: str
```

plugins/modules/nxos_bgp_af.py
praveenramoorthy: Added version

plugins/terminal/nxos.py
```diff
@@ -64,6 +64,14 @@ class TerminalModule(TerminalBase):
             br"Cannot apply ACL to an interface that is a port-channel member",
             re.I,
         ),
+        re.compile(
```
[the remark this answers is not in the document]
praveenramoorthy: Made changes as suggested

praveenramoorthy:
There was problem with rebase, will be opening a new PR with correct changes

PR #179, opened 2020-10-28
Fix vlan config diff to remove default states

rowco:
##### SUMMARY
Fixed an issue with the vlan config generation, where default values were not being properly removed. 

##### ISSUE TYPE
- Bugfix Pull Request

##### ADDITIONAL INFORMATION
When making changes to vlan config default values remained in the commands list. The code removed default values from the existing config, but not from the resulting diff. 

Tested on Nexus 7K platform.

### DIFF
--- a/changelogs/fragments/fix_vlan_diff.yaml
+++ b/changelogs/fragments/fix_vlan_diff.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+  - Vlan config diff was not removing default values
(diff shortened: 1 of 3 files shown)


NilashishC:
@rowco Thank you for the PR. Could you please add an unit test with a scenario where this patch is needed?

rowco:
@NilashishC - I've added a basic test - hope it's ok. The test checks that the module works in an idempotent way for merged/overridden/replaced. Before the bug fix, it would have failed for overridden/replaced.

NilashishC:
@rowco Thank you. The changes look good to me. I see some sanity test and linter failures. We'd need to fix those for this PR to get merged. Could you please run the following commands and push the resulting changes:

```
pip install -r test-requirements.txt
black -l 79 .

Verify linters: tox -elinters .
Verify sanity test: ansible-test sanity --python 3.6 --requirements
```

rowco:
Hi. Ran the commands requested, Black reformatted the dict definition. Hopefully will pass tests now.

NilashishC (approved):

PR #214, opened 2020-12-21
Support vlan interface in static routes.

guoliangdai-personal:
##### SUMMARY

Support vlan interface in static routes like below:
```
ip route 0.0.0.0/0 Vlan10 xxx
```

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

static_routes

##### ADDITIONAL INFORMATION

```
Currently the static_routes module would skip Vlan interface in static routes, so every time we run the commands in this case, it would generate diff and re add the routes even the before and after is same.
```

### DIFF
--- a/changelogs/fragments/205-support-vlan-interface-in-static-routes.yaml
+++ b/changelogs/fragments/205-support-vlan-interface-in-static-routes.yaml
@@ -1,0 +1,3 @@
+---
+doc_changes:
+  - Fix static_routes module skip vlan interface bug.
--- a/plugins/module_utils/network/nxos/facts/static_routes/static_routes.py
+++ b/plugins/module_utils/network/nxos/facts/static_routes/static_routes.py
@@ -122,9 +122,9 @@ class Static_routesFacts(object):
 
         # ethernet1/2/23
         iface = re.match(
-            r".* (Ethernet|loopback|mgmt|port\-channel)(\S*) .*", conf
+            r".* (Ethernet|Vlan|loopback|mgmt|port\-channel)(\S*) .*", conf
         )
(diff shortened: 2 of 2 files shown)


NilashishC:
recheck

NilashishC:
recheck

NilashishC:
Hi @guoliangdai-personal. Thanks a lot for this pull request. Could you add an unit test to cover this scenario as well?

NilashishC (changes requested):
Hi @guoliangdai-personal. Thanks a lot for this pull request. Could you add an unit test to cover this scenario as well?

plugins/module_utils/network/nxos/facts/static_routes/static_routes.py
```diff
@@ -122,9 +122,9 @@ def get_inner_dict(self, conf, inner_dict):
 
         # ethernet1/2/23
         iface = re.match(
-            r".* (Ethernet|loopback|mgmt|port\-channel)(\S*) .*", conf
+            r".* (Ethernet|Vlan|loopback|mgmt|port\-channel)(\S*) .*", conf
```
NilashishC, line 125: The following change is needed allow Vlan interfaces to show up.
```suggestion
            r".* (Ethernet|Vlan|loopback|mgmt|port\-channel)(.*).*", conf
```

NilashishC (changes requested):

changelogs/fragments/205-support-vlan-interface-in-static-routes.yaml
```diff
@@ -0,0 +1,3 @@
+---
+doc_changes:
```
NilashishC, line 2: This would be `bug_fixes`.

Issue #203, opened 2020-11-30
nxos_snmp_community module incorrectly skips configuring a new community string when it sees a partial match


greenjoy3001:
##### SUMMARY
The nxos_snmp_community module is incorrectly skipping configuration of a new community string when the new community string partially matches an existing community string.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
nxos_snmp_community

##### ANSIBLE VERSION

```
ansible 2.9.4
  config file = /root/ansible/test/ansible.cfg
  configured module search path = [u'/root/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.7/dist-packages/ansible
  executable location = /usr/bin/ansible
  python version = 2.7.17 (default, Nov  7 2019, 10:07:09) [GCC 7.4.0]

```

##### CONFIGURATION

```
DEFAULT_GATHERING(/root/ansible/test/ansible.cfg) = explicit
DEFAULT_HOST_LIST(/root/ansible/test/ansible.cfg) = [u'/root/ansible/test/hosts']
DEFAULT_LOG_PATH(/root/ansible/test/ansible.cfg) = /root/log
DEPRECATION_WARNINGS(/root/ansible/test/ansible.cfg) = False
RETRY_FILES_ENABLED(/root/ansible/test/ansible.cfg) = True

```

##### OS / ENVIRONMENT

Nexus device emulated using Cisco NX-OSv9K VIRL image

##### STEPS TO REPRODUCE

```
- hosts: all
  gather_facts: false
  vars: 
    ansible_connection: network_cli
    ansible_network_os: nxos

  tasks: 
      - name: NXOS - Add community test1234
        nxos_snmp_community:
          community: test1234
          access: ro
          state: present

      - name: NXOS - Add community test1
        nxos_snmp_community:
          community: test1
          access: ro
          state: present 
```

##### EXPECTED RESULTS

I would expect two community strings to be configured; test1234 and test1.

The module's idempotency check against the current configuration should check the entire community string, from beginning to end - it appears to be doing a partial match and is incorrectly concluding that the community string already exists. 

##### ACTUAL RESULTS

The first task of the playbook which is to configure community string test1234 executes and returns a status of changed. The second task, which is to configure community string test1 returns a status of ok - at the time of execution, this community string was not configured on the device.

```
TASK [NXOS - Add community test1234] ***************************************************************
changed: [R12]

[...]

PR #230, opened 2021-01-20
nxos_snmp_community: properly handle partial matches in community string

NilashishC:

##### SUMMARY
- Fixes https://github.com/ansible-collections/cisco.nxos/issues/203

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
nxos_snmp_community.py

### DIFF
--- a/changelogs/fragments/nxos_snmp_community.yaml
+++ b/changelogs/fragments/nxos_snmp_community.yaml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - Properly handle partial matches in community string (https://github.com/ansible-collections/cisco.nxos/issues/203).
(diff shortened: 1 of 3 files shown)


GomathiselviS:
recheck

NilashishC:
recheck

NilashishC:
recheck

NilashishC:
recheck

GomathiselviS (commented):
Since we agreed upon increasing the coverage with unit tests, do you think , if  testing this fix with a unit testcase is possible?

NilashishC:
@GomathiselviS This particular case cannot be verified through unit tests. The code change in this patch is related to the on-box "grep" functionality of the target device. `grep word-exp` ensures that we always fetch snmp communities that are an exact match, which is not the case with only `grep`.

GomathiselviS (approved):

PR #244, opened 2021-02-08
Fail gracefully when BGP is already configured with a different ASN

NilashishC:
##### SUMMARY
- Fail gracefully when BGP is already configured with a different ASN when states merged or replaced is used

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
nxos_bgp_global.py


NilashishC:
recheck

justjais (approved):

PR #261, opened 2021-03-05
Fix ternary operator handling to support jinja2 2.10.x and sanity fixes

NilashishC:
Depends-On:  https://github.com/ansible/project-config/pull/762


GomathiselviS:
recheck

NilashishC:
recheck

plugins/module_utils/network/nxos/argspec/bgp_global/bgp_global.py
```diff
@@ -204,9 +204,10 @@ class Bgp_globalArgs(object):  # pylint: disable=R0903
                         },
                         "password": {
                             "type": "dict",
+                            "no_log": False,
```
GomathiselviS, line 207: I think no_log = True will make sense here.

GomathiselviS (changes requested):
no_log should be set to True for keys with sensitive information.

plugins/module_utils/network/nxos/argspec/bgp_global/bgp_global.py
NilashishC: I remember when we discussed this the last time, we concluded that what is sensitive and what is not is subjective to the end user. And as such, a toggle that let's the user sacrifice round trip for security would come in handy. For modules that are already released, adding no_log: True would be a breaking change. I think we need to discuss this again. Will keep the PR open till then.

ashwini-mhatre (approved):

NilashishC:
> no_log should be set to True for keys with sensitive information.

Resolved.

GomathiselviS (approved):

NilashishC:
recheck

PR #459, opened 2022-04-17
Github action for running `ansible-test sanity` and `ansible-test units`

cidrblock:
Enable unit and sanity tests with Github actions.

- Added the GH workflow
- Add a pyproject.toml to be used during dev and CI for consistent results
- Added pytest-ansible-units to test-requirements

Once merged, the "all-green" test should be added to the main branch protection and the zuul unit, sanity and changelog jobs deleted.


.github/workflows/test.yml
```diff
@@ -0,0 +1,34 @@
...
+            "ansible-version": "stable-2.9",
+            "python-version": "3.8"
+          },
+          {
+            "ansible-version": "stable-2.9",
+            "python-version": "3.9"
+          },
+          {
+            "ansible-version": "stable-2.9",
+            "python-version": "3.10"
```
Qalthos, line 32: I'm _pretty_ sure this should be sufficient as we are exclufing 2.9 across all Python versions
```suggestion
            "ansible-version": "stable-2.9",
```

cidrblock: trying it out

NilashishC (approved):

cidrblock:
Recheck

PR #491, opened 2022-05-08
Replace conditional block with an include_tasks

cidrblock:
This will eliminate a skip task message for each task in the block, only the skipped include will be logged.

This should reduce the stdout by a couple thousand lines.

<s>

Seemingly out-of-scope formatting changes are a result of format on save with prettier, and should not happen in the future when #489 is merged.

#489 needs to go first to unblock the pre-commit error related to the collection prep version
</s>


NilashishC:
recheck

NilashishC (approved):

PR #502, opened 2022-05-12
Fix file copy test validation message

NilashishC:


NilashishC:
recheck

PR #504, opened 2022-05-12
Add name to set_fact where missing & update to FQCN

cidrblock:
- Update from `set_fact` to `ansible.builtin.set_fact`
- Provide a name for the test describing the only the root keys being set if missing


cidrblock:
recheck

cidrblock:
Recheck

NilashishC (approved):

PR #513, opened 2022-05-18
Add .virl and collections to gitignore

cidrblock:
These two entries are necessary as a result of the new testing strategy.

- The collection folder is created with linked contents by the pytest-ansible-units plugin
- The .virl directory is a result of using virl within the source root


cidrblock:
I removed the "Restrict who can push to matching branches" prior to adding and removing the mergit label

PR #525, opened 2022-06-07
Fix two ansible-lint rules

ssbarnea:
Related: https://github.com/ansible/ansible-lint/pull/2197


cidrblock (approved):
we might not know if the 0.0.0 breaks a release until later, but it can be on done/ done differently if need be

PR #360, opened 2021-09-16
[ospf_interfaces] Fix rendering of interface names

NilashishC:
Signed-off-by: NilashishC <nilashishchakraborty8@gmail.com>

##### SUMMARY
- Sorting the interfaces with names results in the following:
```
after:
   - name: Ethernet1/1
   - name: Ethernet1/10
   - name: Ethernet1/11
   - name: Ethernet1/12
   - name: Ethernet1/13
   - name: Ethernet1/14
   - name: Ethernet1/2
   - name: Ethernet1/3
   - name: Ethernet1/4
```

- This patch makes it render in the natural order.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
nxos_ospf_interfaces.py

PR #543, opened 2022-07-29
fix OSPF interface related modules failing with subinterfaces

takishida:
##### SUMMARY
When the target NXOS device has sub-interfaces, some modules like `nxos_ospf_interfaces` and `nxos_facts`, which use fact `ospf_interfaces`, fail with the following error:
```
  msg: '''<'' not supported between instances of ''str'' and ''int'''
```

If the target NXOS device has sub-interfaces, these modules fail even when sub-interfaces are not mentioned in your playbook.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
`plugins/module_utils/network/nxos/facts/ospf_interfaces/ospf_interfaces.py`

##### ADDITIONAL INFORMATION
With PR #360 , the sorting of interfaces retrieved as a part of OSPF interfaces fact got improved. However, it introduced a bug where any modules using the said fact under the hood fail when the target device has sub-interfaces.

After the fix in this PR, fact `ospf_interfaces` won't fail even when the target NXOS device has sub-interfaces like below:
```
    ansible_network_resources:
      ospf_interfaces:
      - name: Ethernet1/1
      - name: Ethernet1/1.7
      - name: Ethernet1/2
      - name: Ethernet1/2.30
      - name: Ethernet1/2.31
```

### DIFF
--- a/changelogs/fragments/ospf_subinterface_failure.yaml
+++ b/changelogs/fragments/ospf_subinterface_failure.yaml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - Modules related to OSPF interfaces fail when the target NXOS device has subinterfaces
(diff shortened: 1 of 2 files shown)


NilashishC:
recheck

NilashishC (approved):

PR #544, opened 2022-08-01
Fix var-spacing rule from ansible-lint

ssbarnea:
##### SUMMARY

Reference: https://ansible-lint.readthedocs.io/en/latest/default_rules/#var-spacing

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/tests/integration/targets/nxos_l3_interfaces/tasks/main.yaml
+++ b/tests/integration/targets/nxos_l3_interfaces/tasks/main.yaml
@@ -1,7 +1,7 @@
 ---
 - name: Set a fact for 'rsvd_intf'
   ansible.builtin.set_fact:
-    rsvd_intf: "{{ rsvd_intf|default('mgmt0') }}"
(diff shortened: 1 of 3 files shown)


NilashishC:
recheck

NilashishC (changes requested):
@ssbarnea Thanks for the PR! Could you please add a trivial changelog to this so that CI can pass?

ssbarnea:
TBH, I do not think that this kind of change needs a changelog and I do not know how to tell it that changelog is not needed.

@cidrblock WDYT about that? Any idea if there is a special label or keyword we can add to description to disable that feature? -- similar to `skip-changelog` on devtools projects?

cidrblock:
I think this would do exactly that: https://github.com/ansible-network/github_actions/pull/18  Can you give it a quick review?

ssbarnea:
@NilashishC Sorted! Now we can use the label to tell it to skip requiring a changelog.

NilashishC (approved):

PR #646, opened 2023-03-17
Add route map next hop support

tnizharadze:
##### SUMMARY

This patch adds support for policy based routing and all other stuff under route-map command "set ip next-hop"

Tested on nxos 9000v ver. 9.3.10

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

nxos_route_map

##### ADDITIONAL INFORMATION

```paste below

```


plugins/modules/nxos_route_maps.py
```diff
@@ -332,6 +332,51 @@
                   precedence:
                     description: Set precedence field.
                     type: str
+                  next-hop:
```
NilashishC, line 335:
```suggestion
                  next_hop:
```

plugins/modules/nxos_route_maps.py
```diff
@@ -332,6 +332,51 @@
...
+                      drop_on_fail:
+                        description: Drop packets instead of using default routing when the configured next hop becomes unreachable
+                        type: bool
+                      force_order:
+                        description: Enable next-hop ordering as specified in the address parameter.
+                        type: bool
+                      load_share:
+                        description: Enable traffic load balancing across a maximum of 32 next-hop addresses
+                        type: bool
+                      peer_address: BGP prefix next hop is set to the peer’s local address. If no next hop is set in the route map, the next hop is set to the one stored in the path.
```
NilashishC, line 351: The reason why pre-commit is failing (and also removing some of the doc RSTs) is because of how description is added here. The description needs to assigned to the `description` key.

plugins/modules/nxos_route_maps.py
```diff
@@ -332,6 +332,51 @@
                   precedence:
                     description: Set precedence field.
                     type: str
+                  next_hop:
```
NilashishC, line 335: Could you please add (a) unit tests, (b) an example to show how the newly added key works? Thank you.

NilashishC (changes requested):

plugins/modules/nxos_route_maps.py
tnizharadze: It`s OK, I found solution.

tnizharadze: Done. But during tests I found some anomalies in NXOS (ver 9.3(10)). If we enter these commands in switch:
`route-map TEST permit 10`
`set ip next-hop verify-availability 3.3.3.3 track 1 drop-on-fail force-order load-share`
`set ip next-hop verify-availability 5.5.5.5 track 3 drop-on-fail force-order load-share`
then 'show run | sec route-map' produces these output:
`route-map TEST permit 10`
`  set ip next-hop verify-availability 3.3.3.3 track 1`
`  set ip next-hop verify-availability 5.5.5.5 track 3 load-share force-order drop-on-fail`
Provided test and examples does not cover this situation.
From ansible point of view this issue means that in that situation just idempotensy will not work properly.

plugins/modules/nxos_route_maps.py
tnizharadze: done

NilashishC (approved):

PR #667, opened 2023-04-23
Run GHA periodically

NilashishC:
##### SUMMARY
- Run GHA workflows at 00:00 UTC every day.
- This is an attempt to replace Zuul's periodic pipeline jobs. But this unfortunately, has the following caveat: https://upptime.js.org/blog/2021/01/22/github-actions-schedule-not-working/


NilashishC:
recheck

Qalthos (approved):

Issue #370, opened 2021-09-29
Create NX-OS user providing a hashed password


DavidMorp:
##### SUMMARY
Add the possibility to configure hashed passwords on Cisco NX-OS devices.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
hashed_password

##### ADDITIONAL INFORMATION
Please keep the same syntax as the IOS component: https://docs.ansible.com/ansible/latest/collections/cisco/ios/ios_user_module.html#parameter-hashed_password
```yaml
- name: Add a user with MD5 hashed password
  cisco.ios.ios_user:
    name: ansibletest5
    hashed_password:
      type: 5
      value: $3$8JcDilcYgFZi.yz4ApaqkHG2.8/
```

aalmenar:
This could be a great addition to the nxos_user module

NilashishC:
This has been added with #677. Hence, closing this ticket.

PR #677, opened 2023-05-15
[nxos_user] add hashed_password option

ashleytayles:
##### SUMMARY
Adds a 'hashed_password' argument to nxos_user to allow user to provide an already hashed password.
 
Closes:
#370 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
nxos_user

##### ADDITIONAL INFORMATION
Used the same "hashed_password" key name for continuity from cisco.ios collection. However I chose to KISS and make it a string rather than a dict containing 'type' and 'value', as NX-OS does not have multiple encryption algorithms to chose from unlike IOS.
'hashed_password' and 'configured_password' set as mutually exclusive.
Tests provided value with a quick regex to try and validate a hashed password has been provided.

Tested on Nexus 9300v
NXOS: version 10.1(1)

### DIFF
--- a/plugins/modules/nxos_user.py
+++ b/plugins/modules/nxos_user.py
@@ -57,6 +57,11 @@ options:
         - The password to be configured on the network device. The password needs to be
           provided in cleartext and it will be encrypted on the device.
         type: str
+      hashed_password:
+        description:
+        - The hashed password to be configured on the network device. The password needs to 
+          already be encrypted.
+        type: str
       update_password:
         description:
         - Since passwords are encrypted in the device running config, this argument will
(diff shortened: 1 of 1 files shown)


NilashishC:
@ashleytayles Thank for so much for this PR! Could you also please add an unit test case for the new option?

NilashishC (changes requested):
@ashleytayles Thank for so much for this PR! Could you also please add an unit test case for the new option?

ashleytayles:
> @ashleytayles Thank for so much for this PR! Could you also please add an unit test case for the new option?

Absolutely! Committing soon..

NilashishC (approved):

PR #717, opened 2023-07-28
Fix var-naming[no-reserved] lint errors

NilashishC:


KB-perByte (approved):

NilashishC:
The `all_green` check fails because it expects all workflows to end with `success` and does not honour `skipped`, which is what resulted for the changelog test.

KB-perByte:
This to be updated https://github.com/ansible-collections/cisco.ios/blob/def8fe19c94c37949c7a518e49dee11c72166456/.github/workflows/tests.yml#L49

NilashishC:
recheck

NilashishC:
recheck

PR #738, opened 2023-08-25
fc_interfaces module for cisco mds

srbharadwaj:
##### SUMMARY
This PR is for fc_interfaces module for cisco mds

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
`fc_interfaces`


srbharadwaj (commented):
auto-fixes

NilashishC (changes requested):
This PR needs a changelog entry similar to https://github.com/ansible-collections/cisco.nxos/blob/main/changelogs/fragments/_health_check_update_docs.yaml#L2.

The section will be `minor_changes` instead of  `doc_changes`.

NilashishC (changes requested):
Please fix the sanity issues reported by CI.

srbharadwaj:
@NilashishC  
can u please tell me how to fix this?
https://github.com/ansible-collections/cisco.nxos/actions/runs/5973688660/job/16206411773?pr=738
and this?
https://github.com/ansible-collections/cisco.nxos/actions/runs/5973688660/job/16206417490?pr=738

changelogs/fragments/_health_check_update_docs.yaml
```diff
@@ -1,9 +1,3 @@
 ---
-doc_changes:
-  - Update examples for ospfv2 resource modules using yaml callback plugin.
-  - Update examples for ospfv3 resource modules using yaml callback plugin.
-  - Update examples for ospf_interfaces resource modules using yaml callback plugin.
-  - Update examples for bgp_global resource modules using yaml callback plugin.
-  - Update examples for bgp_address_family resource modules using yaml callback plugin.
-  - Update examples for bgp_templates resource modules using yaml callback plugin.
-  - Update examples for bgp_neighbor_address_family resource modules using yaml callback plugin.
+minor_changes:
```
NilashishC, line 2: Looks like you edited an existing changelog fragment. You need to add a new one for your changes.

plugins/modules/nxos_fc_interfaces.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+__metaclass__ = type
+
+DOCUMENTATION = """
+---
+module: nxos_fc_interfaces
+short_description: Fc Interfaces resource module
+description: This module manages the interface attributes of NX-OS fc interfaces.
+version_added: 1.0.0
```
NilashishC, line 21:
```suggestion
version_added: 6.0.0
```

plugins/modules/nxos_fc_interfaces.py
```diff
@@ -0,0 +1,201 @@
...
+          - Port mode of the fc interface
+        choices:
+          - auto
+          - E
+          - F
+          - Fx
+          - NP
+          - SD
+        type: str
+      trunk_mode:
```
NilashishC, line 86: Error from sanity test:
`ERROR: plugins/modules/nxos_fc_interfaces.py:0:0: doc-choices-do-not-match-spec: Argument 'trunk_mode' in argument_spec found in config defines choices as (['auto', 'on', 'off']) but documentation defines choices as (['auto', 'True', 'False'])`

You either need to update the doc or the argspec so that the choices for this option are identical.

plugins/modules/nxos_fc_interfaces.py
```diff
@@ -0,0 +1,201 @@
...
+      - merged
+      - replaced
+      - overridden
+      - deleted
+      - gathered
+      - rendered
+      - parsed
+    default: merged
+"""
+EXAMPLES = """
```
NilashishC, line 116: Please add some examples in the following format:

https://github.com/ansible-collections/cisco.nxos/blob/main/plugins/modules/nxos_interfaces.py#L128

plugins/modules/nxos_fc_interfaces.py
```diff
@@ -0,0 +1,201 @@
...
+      description:
+        description:
+          - Interface description.
+        type: str
+      enabled:
+        description:
+          - Administrative state of the interface. Set the value to C(true) to
+            administratively enable the interface or C(false) to disable it
+        type: bool
+      speed:
```
NilashishC, line 55: Error from sanity test:

`ERROR: plugins/modules/nxos_fc_interfaces.py:0:0: doc-choices-incompatible-type: DOCUMENTATION.options.config.suboptions.speed: Argument defines choices as (1000) but this is incompatible with argument type str: Value must be string for dictionary value @ data['options']['config']['suboptions']['speed']. Got {'description': ['Interface link speed.'], 'choices': ['auto', 1000, 2000, 4000, 8000, 10000, 16000, 32000, 64000, 'auto max 2000', 'auto max 4000', 'auto max 8000', 'auto max 16000', 'auto max 32000', 'auto max 64000'], 'type': 'str'}`

I think this can be fixed if you put quotes around the integer values - '1000', '2000', etc.

plugins/modules/nxos_fc_interfaces.py
NilashishC: Since you have defined this type as `str`, I would recommend keeping `on` and `off`, instead of `True` and `False`.

srbharadwaj: in both places i don't see the boolen values `True/False`, can you point me which file has the bool values?

NilashishC: @srbharadwaj Looks like you need to enclose the options `on` and `off` in quotes. Otherwise the YAML parser seems to interpret them as True/False from the documentation.

plugins/modules/nxos_fc_interfaces.py
```diff
@@ -0,0 +1,201 @@
...
+          - Fx
+          - NP
+          - SD
+        type: str
+      trunk_mode:
+        description:
+          - Trunk mode of the fc interface
+        choices:
+          - auto
+          - on
```
NilashishC, line 91:
```suggestion
          - 'on'
```

plugins/modules/nxos_fc_interfaces.py
```diff
@@ -0,0 +1,201 @@
...
+          - off
```
NilashishC, line 92:
```suggestion
          - 'off'
```

srbharadwaj:
yes will do in a day or two, busy with other things

srbharadwaj:
@NilashishC @KB-perByte let me know if anything else needs to be done from my side

plugins/modules/nxos_fc_interfaces.py
```diff
@@ -0,0 +1,491 @@
...
+#   enabled: False
+
+# After state:
+# ------------
+#
+# switch# show running-config interface all
+# interface fc1/2
+#     switchport speed auto
+#     switchport mode auto
+#     no switchport description
```
NilashishC, line 286: Is this the actual output from the device CLI?

plugins/modules/nxos_fc_interfaces.py
```diff
@@ -0,0 +1,491 @@
...
+#   mode: auto
+#   trunk_mode: on
+#   enabled: True
+#   description: This is a sample line
+# - name: fc1/2
+#   speed: 1000
+#   mode: E
+#   trunk_mode: off
+#   enabled: True
+
```
NilashishC, line 404: Can you please add an example for `overridden`?

plugins/module_utils/network/nxos/rm_templates/fc_interfaces.py
```diff
@@ -0,0 +1,194 @@
...
+            "setval": "analytics type {{ analytics_nvme|string }}",
+            "result": {
+                "{{ name }}": {
+                    "analytics_nvme": "{{ analytics_nvme }}",
+                },
+            },
+        },
+
+        {
+            "name": "analytics",
```
NilashishC, line 179: In which scenario is this parser being used?

plugins/module_utils/network/nxos/facts/fc_interfaces/fc_interfaces.py
```diff
@@ -0,0 +1,103 @@
...
+
+class Fc_interfacesFacts(object):
+    """The nxos fc_interfaces facts class"""
+
+    def __init__(self, module, subspec="config", options="options"):
+        self._module = module
+        self.argument_spec = Fc_interfacesArgs.argument_spec
+
+    def get_interfaces_data(self, connection):
+        return connection.get("show running-config interface all")
```
NilashishC, line 39: One standard convention that we used in all modules was to _not_ include default values while pulling running-config. This let's us ignore those values (while rendering facts) that do not show up in `show running-config | section interface`.

Can you share we are doing this here?

NilashishC: This also makes the `deleted` operation behave differently than all other interfaces modules.

plugins/modules/nxos_fc_interfaces.py
srbharadwaj: yes

plugins/modules/nxos_fc_interfaces.py
srbharadwaj: overrideen and replaced are same what will be the difference?

plugins/module_utils/network/nxos/rm_templates/fc_interfaces.py
srbharadwaj: didnt get you

plugins/modules/nxos_fc_interfaces.py
NilashishC: I'm not sure I get it. Why is overridden and replaced state the same?

plugins/module_utils/network/nxos/rm_templates/fc_interfaces.py
NilashishC: This parser doesn't seem to have a setval. Can you explain that? And also why we need this particular parser when we already have to specific ones above this?

plugins/module_utils/network/nxos/facts/fc_interfaces/fc_interfaces.py
srbharadwaj: if any config are default then it wont show up in show running-config  hence we use show running-config interface all to get the default values

plugins/module_utils/network/nxos/rm_templates/fc_interfaces.py
srbharadwaj: we had discussed this over a call regrading this if you remember, because analytics fc-all shows up differently while getting the facts and while configuring its different

plugins/module_utils/network/nxos/facts/fc_interfaces/fc_interfaces.py
srbharadwaj: deleted for interfaces module means moving to default values isnt it?

plugins/modules/nxos_fc_interfaces.py
srbharadwaj: i think we need a call to over these things

plugins/module_utils/network/nxos/rm_templates/fc_interfaces.py
NilashishC: Oh, looking for closely at the examples, I get that now.

plugins/module_utils/network/nxos/facts/fc_interfaces/fc_interfaces.py
NilashishC: > deleted for interfaces module means moving to default values isnt it?

Yes, but we do not consider the defaults that show up in `running-config all`? Did you have a look at how it is done in the nxos_interfaces module?

plugins/modules/nxos_fc_interfaces.py
NilashishC: Sure, please feel free to send an invite.

srbharadwaj:
@NilashishC added the overridden cases, can you check now?

NilashishC (approved):

PR #772, opened 2023-10-19
Add to_text for vrf names

wormley:
##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

cisco.nxos.nxos_bgp_global
##### ADDITIONAL INFORMATION
The cisco.nxos.nxos_bgp_global suffers the same edge case as was fixed in snmp last year. If there are mixed vrf names some having all digits and some having letters the same error occurs in sorted() when it tries to compare int to string.
This is the same code change as was done in snmp_server.

This proposed change may cause some issue for anyone using purely numeric vrf names as "9" will now sort after "10" and the section will be re-ordered.

### DIFF
--- a/plugins/module_utils/network/nxos/facts/bgp_global/bgp_global.py
+++ b/plugins/module_utils/network/nxos/facts/bgp_global/bgp_global.py
@@ -15,6 +15,7 @@ for a given resource, parsed, and the facts tree is populated
 based on the configuration.
 """
 
+from ansible.module_utils._text import to_text
 from ansible.module_utils.six import iteritems
 from ansible_collections.ansible.netcommon.plugins.module_utils.network.common import utils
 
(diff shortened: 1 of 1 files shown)


KB-perByte:
Hey @wormley as you correctly pointed out _"This proposed change may cause some issue for anyone using purely numeric vrf names as "9" will now sort after "10" and the section will be re-ordered."_

I would recommend if we can have a fresh implementation of a sorting algorithm that can handle both the pattern of vrf names and place it in a generic place like netcommon, it would have been the right approach to this issue.

Let us know if you need any information on this.
Regards.

NilashishC (changes requested):
This would require implementing the changes suggested by @KB-perByte, before we can take it forward. Please let us know if you need any help on this.

Thank you for your contribution! We really appreciate it.

PR #849, opened 2024-04-11
CML LAB unique name

roverflow:
##### SUMMARY
- Mostly changing the single.yml file to have title as collection full name so it can find and replace later using sed
- Passing the labtitle prefix in integration test


roverflow:
> I'm unsure what these changes are doing and why is it required? Can you please elaborate?

Sure, just updated the summary

NilashishC (approved):

PR #866, opened 2024-06-06
Bump min `ansible-core` version to `2.15.0`

NilashishC:


Ruchip16 (approved):

changelogs/fragments/bump_215.yaml
```diff
@@ -0,0 +1,6 @@
+---
+release_summary: >
```
Ruchip16, line 2: we dont need release_summary here, it would be required in the maor release PR thats why changelog CI action is failing

PR #876, opened 2024-07-31
Tests/Fixes the device not entering vrf context

roverflow:
##### SUMMARY
Tests/Fixes the device not entering vrf context

##### ISSUE TYPE

- Test fix


NilashishC (approved):
The changes look good to me. Please merge if CI passes both upstream and downstream.

PR #877, opened 2024-08-02
[nxapi] Add `vrf` option to `nxos_nxapi` module

Miyoshi-Ryota:
##### SUMMARY

Add `vrf` option to `nxos_nxapi` module.

###### Comments on the design:
It's an old module with only state: present / absent. As far as I can see, when state: present is specified, it does not change parameters that are not mentioned. For example, if the `nxapi sandbox` command is already present on the device but the sandbox parameter is not specified in the playbook, it results that no deletion or other actions are taken.

Therefore, the vrf option follows this design.

Not applying the `nxapi use-vrf` command at all means enabling nxapi for all VRFs. However, in my implementation of this module, there is currently no way to remove an already applied nxapi use-vrf <xxxxx> configuration at all. This is because while it is possible to specify another VRF, but there is no way to reflect the absence of a parameter. This aligns with the aforementioned behavior of state: `present` where nothing is done if a parameter is not specified.

This implementation method was chosen to maintain consistency within the module, avoid disrupting environments using the existing nxos_nxapi module, and keep the parameters simple.

Other possible approaches:
- Implementing a behavior similar to the state: overriden in other modules, meaning if no VRF is specified, the existing nxapi use-vrf setting would be deleted.
- Making the parameter more complex. For example, explicitly enabling or disabling as follows:
```
vrf:
  limit_to_vrf: bool
  vrf_name: xxx
```

If you prefer to select one of other two approaches I mentioned, or an entirely different approach, please let me know.  I am willing to make adjustments accordingly.

###### Reference for the supported versions of use-vrf:
- For N9K: 
  - At least, supported in version 7.0(3)I3(1)
    - https://www.cisco.com/c/en/us/td/docs/switches/datacenter/nexus9000/sw/7-x/command_references/configuration_commands/b_Using_N9K_Config_Commands/b_N9K_Bookmap_chapter_01111.html#wp4568333650 
  - Also, it is not supported in version 6.1(2)I2(2)
    - https://www.cisco.com/c/en/us/td/docs/switches/datacenter/nexus9000/sw/6-x/command_reference/config_612I22/b_n9k_command_ref/b_n9k_command_ref_chapter_010000.html#wp2897910537
  - I am unsure of the exact version where it was released. I have reviewed all the release notes, but there is no mention of it.
- For N7K:
  - > The nxapi use-vrf feature is introduced in Cisco NX-OS Release 8.2(3), which helps to secure NX-API by binding the NX-API to a particular VRF.
  
    - https://www.cisco.com/c/en/us/td/docs/switches/datacenter/nexus7000/sw/programmability/guide/cisco_nexus7000_programmability_guide_8x/b-cisco-nexus7000-programmability-guide-8x_chapter_011.html

- For N3K:
  - At least, supported in version 7.0(3)I4(1)
    - https://www.cisco.com/c/en/us/td/docs/switches/datacenter/nexus3000/sw/command/reference/703i74m3k/config/b_N3K_Config_Commands_703i74/b_N3K_Config_Commands_703i74_chapter_01111.html#wp1351653258

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

nxos_nxapi_module

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/module_utils/network/nxos/utils/utils.py
+++ b/plugins/module_utils/network/nxos/utils/utils.py
@@ -189,7 +189,7 @@ class Version:
     """Simple class to compare arbitrary versions"""
 
     def __init__(self, version_string):
-        self.components = version_string.split(".")
+        self.components = list(map(int, version_string.split(".")))
 
     def __eq__(self, other):
         other = _coerce(other)
--- a/changelogs/fragments/nxos_nxapi.yaml
+++ b/changelogs/fragments/nxos_nxapi.yaml
@@ -1,3 +1,4 @@
 ---
 minor_changes:
   - "nxos_nxapi - Add vrf parameter to nxos_nxapi module"
+  - "nxos_nxapi - Fix ssl_strong_ciphers option is not working over version 10"
--- a/changelogs/fragments/nxos_nxapi.yaml
+++ b/changelogs/fragments/nxos_nxapi.yaml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - "nxos_nxapi - Add vrf parameter to nxos_nxapi module"
--- a/changelogs/fragments/nxos_nxapi.yaml
+++ b/changelogs/fragments/nxos_nxapi.yaml
@@ -1,4 +1,5 @@
 ---
 minor_changes:
   - "nxos_nxapi - Add vrf parameter to nxos_nxapi module"
+bugfixes:
   - "nxos_nxapi - Fix ssl_strong_ciphers option is not working over version 10"
--- a/docs/cisco.nxos.nxos_nxapi_module.rst
+++ b/docs/cisco.nxos.nxos_nxapi_module.rst
@@ -232,7 +232,7 @@ Parameters
                 <td>
                 </td>
                 <td>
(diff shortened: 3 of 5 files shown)


plugins/module_utils/network/nxos/utils/utils.py
```diff
@@ -189,7 +189,7 @@ class Version:
     """Simple class to compare arbitrary versions"""
 
     def __init__(self, version_string):
-        self.components = version_string.split(".")
+        self.components = list(map(int, version_string.split(".")))
```
Miyoshi-Ryota, line 192: When each part of the version was of type str, it does not work correctly once the version exceeds 10. So We need to convert to str to int.

```
>>> ["10", "3"] > ["9", "1"]
False
>>> "10" > "9"
False
>>>
```

Miyoshi-Ryota:
All my work is completed and it is ready for review.

If this commit needs to be split into a separate PR, please let me know.
https://github.com/ansible-collections/cisco.nxos/pull/877/commits/3cb4fef5bb939c584aef49db8774c5f9d4a47f24

Miyoshi-Ryota:
All my work is completed and it is ready for review.

PR #880, opened 2024-08-07
Fixes Resource module cisco.nxos.nxos_acls is not correctly converting the ports under port_protocol range keys

roverflow:
##### SUMMARY
Fixes - [ANA-520](https://issues.redhat.com/browse/ANA-520)
@sean-m-sullivan

Fixes Resource module cisco.nxos.nxos_acls is not correctly converting the ports under port_protocol range keys

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -46,11 +46,11 @@ repos:
         args: ["--filter-files"]
 
   - repo: https://github.com/psf/black
-    rev: 24.4.2
+    rev: 24.8.0
     hooks:
       - id: black
 
   - repo: https://github.com/pycqa/flake8
-    rev: 7.1.0
+    rev: 7.1.1
     hooks:
       - id: flake8


changelogs/fragments/bugfix_vrf_range_resolution.yml
```diff
@@ -0,0 +1,3 @@
+---
+bugfixes:
+  - "acls - fix lookup of range port conversion from int to string to allow strings. (https://github.com/ansible-collections/cisco.nxos/issues/???)."
```
KB-perByte, line 3: Make sure we update the issue link

KB-perByte, line 3:
```suggestion
  - "acls - Fix lookup of range port conversion from int to string to allow strings. (https://github.com/ansible-collections/cisco.nxos/issues/???)."
```

sean-m-sullivan: @KB-perByte I've updated this to point to redhat, or should we point somewhere else

PR #894, opened 2024-09-03
[nxos_prefix_lists] Fix idempotency issues caused by IP prefix notation discrepancies

Miyoshi-Ryota:
##### SUMMARY

Fixed an issue where idempotency was not maintained due to discrepancies in the notation of IP prefixes. By introducing the `ipaddress` module to standardize prefixes, we ensure that the playbook and running-config prefixes are compared accurately, preventing unnecessary configuration changes.

For example, previously, a playbook specifying `prefix: ::0/` and a running-config showing `ipv6 prefix-list plist1 seq 10 permit 0::/0`, or a playbook specifying `prefix: 10.0.0.8/24` and a running-config showing `ip prefix-list plist2 seq 10 permit 10.0.0.0/24`, would result in unnecessary changes. With this update, such discrepancies are handled, maintaining idempotency and ensuring existing configurations are not repeatedly deleted and re-added.

Supplementary Information:
The notation you enter in configure terminal and the notation in the `show run` output may differ on Nexus devices. Therefore, a simple text comparison could break idempotency.

An example is shown below.
```
# conf t
(config)# ipv6 prefix-list plist1 seq 10 permit ::/0
(config)# ip prefix-list plist2 seq 10 permit 10.0.0.8/24
(config)# exit
# show run | inc plist 
ip prefix-list plist2 seq 10 permit 10.0.0.0/24
ipv6 prefix-list plist1 seq 10 permit 0::/0
```

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

nxos_prefix_lists

##### ADDITIONAL INFORMATION

Before this change, when I executed the playbook below multiple times, the result would always be changed, and plist1 would be deleted and added each time.

```paste below
- hosts: all
  gather_facts: no
  vars:
    ansible_connection: network_cli
    ansible_user: "hoo"
    ansible_password: "bar"
    ansible_network_os: cisco.nxos.nxos
  tasks:
    - name: prefixlists_always_changed
      cisco.nxos.nxos_prefix_lists:
        state: "overridden"
        config:
          - afi: ipv6
            prefix_lists:
            - entries:
              - action: permit
                prefix: ::0/0
                sequence: 10
              name: "plist1"
```

### DIFF
--- a/changelogs/fragments/nxos_prefix_lists.yaml
+++ b/changelogs/fragments/nxos_prefix_lists.yaml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - "nxos_prefix_lists - Fix idempotency issues caused by IP prefix notation discrepancies"
--- a/plugins/module_utils/network/nxos/config/prefix_lists/prefix_lists.py
+++ b/plugins/module_utils/network/nxos/config/prefix_lists/prefix_lists.py
@@ -18,6 +18,14 @@ necessary to bring the current configuration to its desired end-state is
 created.
 """
 
+try:
+    import ipaddress
+    HAS_IPADDRESS = True
+except ImportError:
+    HAS_IPADDRESS = False
+
+import copy
+
 from ansible.module_utils.six import iteritems
 from ansible_collections.ansible.netcommon.plugins.module_utils.network.common.rm_base.resource_module import (
     ResourceModule,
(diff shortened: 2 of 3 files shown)


Miyoshi-Ryota:
@Ruchip16 Hello, what do you think about this pull request?

I believe that the fixes in this PR may not hold much value for IPv4, but they do for IPv6. In IPv6, there are many correct representations in principle. Additionally, Cisco's notation does not always seem to be the most common one.

For example, when I ask network engineers around me, they agree, and considering the existence of RFC 5952, ::/0 seems to be a more standard representation than 0::/0, which is required by Cisco and  current this module.

Therefore, in its current state, this module might lead to issues unless one has a deep understanding of Ansible.

Hence, I believe this PR has certain value.

Issue #871, opened 2024-06-20
cisco.nxos.nxos_user - Purge deletes users which does not exist, breaks idempotency


jorgenspange:
For fetching existing users the module runs "show user-account", this command displays users which does not exist, only have been logged in through SSH:

```yaml
user:sshuser
        roles:network-admin vdc-admin
account created through REMOTE authentication
Credentials such as ssh server key will be cached temporarily only for this user account
Local login not possible
```

This makes the module try to delete these which are breaking idempotency as they are not present anywhere in the configuration.

The module also deletes users which is defined for snmp, which is more understandable, but still unfortunate.

AAYUSH2091:
Hi @jorgenspange ,

After going through your issue, I see you mentioned that logged-in users are being deleted. If these users are local, shouldn't they have an entry in the local user database, meaning the purge function is working as expected? Could you please explain what behavior you expect from the purge process? 

and also can you share the playbook as well that you were using.

jorgenspange:
Hi @AAYUSH2091,

Sorry for the misunderstanding, the logged in users i mention is tacacs-users and not locally created ones.
This happens because "show user-account" also displays ssh connected accounts which is authenticated with tacacs.

this is my play:

```yaml
- name: remove all users except admin
  cisco.nxos.nxos_user:
    purge: true
```

Ideally this would delete all locally configured users which are not admin.
What it does is that it tries to delete:
locally configured users
users that are logged in and not defined locally
users which are defined for other purposes for instance snmp

NilashishC:
@jorgenspange Hi! Would it be possible to give @AAYUSH2091's PR #901 a try? Thanks!

PR #903, opened 2024-10-21
Fix nxos_user purge deleting non-locally configured users.

AAYUSH2091:
SUMMARY
Fixes an issue in the nxos_user module where the purge option deleting the non-locally configured users displayed by the "show user-account" command, breaking idempotency. Now, only locally configured users are purged.

Fixes https://github.com/ansible-collections/cisco.nxos/issues/871

ISSUE TYPE
Bugfix Pull Request
COMPONENT NAME
nxos_user

ADDITIONAL INFORMATION
When the purge option was used, users authenticated via TACACS+ or users defined for SNMP were being deleted despite not being locally configured. This fix ensures only local users are purged.


Ruchip16:
@AAYUSH2091 can you please add integration tests to validate the changes?

NilashishC (approved):

PR #913, opened 2024-12-07
bugfix: Fix `no ipv6_redirects command` not being sent from module to device"

Ruchip16:
##### SUMMARY

When an interface is already configured with both an ipv4 address and ipv6 address (dual stacked), the `no ipv6 redirect` command is NOT sent from the module to the device.

##### ISSUE TYPE

- Bugfix Pull Request
- link: [ANA-599](https://issues.redhat.com/browse/ANA-599)

##### COMPONENT NAME

`nxos.l3_interfaces`
##### ADDITIONAL INFORMATION

So the NX-OS module has a built-in behaviour where the device may auto-disable IPv6 redirects in certain scenarios (e.g., dual-stack configuration), and explicitly setting both redirects and ipv6_redirects may be required to achieve the desired state.

- Device Behavior:
The note in the [add_commands](https://github.com/ansible-collections/cisco.nxos/blob/main/plugins/module_utils/network/nxos/config/l3_interfaces/l3_interfaces.py) function, indicates the following: 

> "Device will auto-disable redirects when secondaries are present."

means that the device has a built-in behavior where it auto-disables IPv6 redirects in certain scenarios, such as dual-stacked interfaces (both IPv4 and IPv6 configured) or when multiple IP addresses are present (secondary addresses).
The no ipv6 redirects command may depend on the state of redirects (IPv4 redirects). This is an inherent behavior of the device.

**1. Expected Behavior:**

When ipv6_redirects: false is specified in the playbook, the module should send the command: `no ipv6 redirects`

**2. Actual Behavior: (My observation)**

If only ipv6_redirects: false is set, the no ipv6 redirects command may be skipped because the device already auto-disables IPv6 redirects.

Setting both redirects: false and ipv6_redirects: false ensures both commands (no ip redirects and no ipv6 redirects) are sent, overriding the device's implicit behavior.

**3. Proposed solution 1: Update the playbook**

```
- name: Configure interface
  hosts: nxos
  gather_facts: false
  tasks:
    - name: Disable IPv6 redirects
      cisco.nxos.nxos_l3_interfaces:
        config:
          - name: Ethernet1/2
            redirects: false
            ipv6_redirects: false
        state: merged
```
this would execute the following commands:

```
 "commands": [
        "interface Ethernet1/2",
        "no ip redirects",
        "no ipv6 redirects"
    ],
```
    
**Proposed solution 2**: Update the code to execute the command when explicitly configured in the playbook, regardless of the device's current behavior


tests/unit/modules/network/nxos/test_nxos_l3_interfaces.py
```diff
@@ -1044,3 +1044,26 @@ def test_replaced_tag(self):
         playbook["state"] = "replaced"
         set_module_args(playbook, ignore_provider_arg)
         self.execute_module(changed=True, commands=commands)
+
+    def test_ipv6_redirects(self):
```
NilashishC, line 1048: This test case seems to pass even without the changes in this PR. I don't think this covers the bug being solved.

plugins/module_utils/network/nxos/config/l3_interfaces/l3_interfaces.py
```diff
@@ -482,10 +482,10 @@ def add_commands(self, diff, name=""):
                 self.cmd_order_fixup(commands, name)
         if "ipv6_redirects" in diff:
             # Note: device will auto-disable redirects when secondaries are present
-            if diff["ipv6_redirects"] != self.check_existing(name, "ipv6_redirects"):
-                no_cmd = "no " if diff["ipv6_redirects"] is False else ""
-                commands.append(no_cmd + "ipv6 redirects")
-                self.cmd_order_fixup(commands, name)
+            # if diff["ipv6_redirects"] != self.check_existing(name, "ipv6_redirects"):
```
NilashishC, line 485: This check must have been added previously for a reason and we should not drop it without a proper justification.

PR #916, opened 2024-12-19
Fix handling of "none" in allowed_vlans for nxos_l2_interfaces

AAYUSH2091:
##### SUMMARY
Fixes handling of "none" value in allowed_vlans configuration for nxos_l2_interfaces module. Currently, when configuring allowed_vlans: none with state: replaced, the module either fails with int conversion error or removes the VLAN configuration entirely, making the port allow all VLANs instead of setting it to none.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
cisco.nxos.nxos_l2_interfaces

##### ADDITIONAL INFORMATION
The issue occurs in two scenarios:

1. When directly setting allowed_vlans to "none":
```yaml
- name: Configure VLAN none
  cisco.nxos.nxos_l2_interfaces:
    config:
      - name: Ethernet1/1
        trunk:
          allowed_vlans: "none"
    state: replaced


AAYUSH2091:
closed this Pr as integration test are not passing

PR #922, opened 2025-01-10
Fix downstream tests

roverflow:
##### SUMMARY
Fix downstream tests for vrf_address_family, vrf_interfaces

##### ISSUE TYPE

- Test fix


tests/integration/targets/nxos_vrf_interfaces/tasks/main.yaml
```diff
@@ -1,12 +1,5 @@
...
-        - cli
-  always:
-    - name: Include the NX-API tasks
-      ansible.builtin.include_tasks: nxapi.yaml
-      tags:
-        - nxapi
+- name: Main task for vrf_interfaces module
+  ansible.builtin.include_tasks: cli.yaml
+  tags:
+    - network_cli
```
NilashishC, line 5: Why did we remove NX-API tests for this target?

NilashishC (changes requested):
I think we should wrap all task/main.yml of all such failing target roles with a var, let's say "nxos_skip_marked" and ask @komaldesai13 to set this var to true in downstream inventory.

This change is temporary and I really do not want to touch the existing logic.

tests/integration/targets/nxos_vrf_interfaces/tasks/main.yaml
roverflow: had changed it when local testing! reverted back

tests/integration/targets/nxos_vrf_interfaces/tasks/cli.yaml
```diff
@@ -1,31 +1,26 @@
...
-    paths: "{{ role_path }}/tests/cli"
-    patterns: "{{ testcase }}.yaml"
-  connection: local
-  register: cli_cases
-
-- name: Set a fact for 'test_cases'
-  ansible.builtin.set_fact:
-    test_cases:
-      files: "{{ test_cases.files + cli_cases.files }}"
+  delegate_to: localhost
```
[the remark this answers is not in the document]
roverflow: reverted it back

tests/integration/targets/nxos_vrf_address_family/tasks/cli.yaml
```diff
@@ -13,6 +13,7 @@
   delegate_to: localhost
 
 - name: Run test case (connection=ansible.netcommon.network_cli)
+  when: not nxos_skip_marked | default(false)
```
NilashishC, line 16: If we're not running the tests at all, the above collection tasks shouldn't be run as well. Can't we add this conditional to tasks/main.yml, which is the entry point?

roverflow: Done!

PR #924, opened 2025-01-15
Fix vrf_interfaces test cases

roverflow:
##### SUMMARY
Fix vrf_interfaces test cases due to difference in interfaces

##### ISSUE TYPE
- Tests


tests/integration/targets/nxos_vrf_interfaces/tests/common/gathered.yaml
```diff
@@ -16,7 +16,9 @@
       ansible.builtin.assert:
         that:
           - not result.changed
-          - "{{ gathered['config'] | symmetric_difference(result['gathered']) |length == 0 }}"
+          # Since different devices may have different interfaces, we can't assert the exact number of interfaces
```
NilashishC, line 19: These tests are _only_ supposed to work on a very specific chassis x NX-OS combination - both upstream and downstream. In that case, this shouldn't be an issue.

tests/integration/targets/nxos_vrf_interfaces/tests/common/_gather_interfaces.yaml
```diff
@@ -0,0 +1,10 @@
+---
+- name: Gather the provided configuration
+  register: result
+  cisco.nxos.nxos_vrf_interfaces:
```
NilashishC, line 4: Why are we using nxos_vrf_interfaces in tests for the same module?

tests/integration/targets/nxos_vrf_interfaces/tests/common/gathered.yaml
roverflow: currently when testing, i was having upto 64 interfaces in the downstream device, we would have to add all these 64 interfaces in our vars/main.yaml for every list we create hence im asserting if the change i want exist instead of adding all of them

tests/integration/targets/nxos_vrf_interfaces/tests/common/_gather_interfaces.yaml
roverflow: this is only to gather the before state to asset it, hence using this

tests/integration/targets/nxos_vrf_interfaces/tests/common/gathered.yaml
NilashishC: Does the upstream one have 64 interfaces as well? If so, then we can just check for N interfaces that are touched by this role.

tests/integration/targets/nxos_vrf_interfaces/tests/common/_gather_interfaces.yaml
NilashishC: This is not a good approach, IMO. I know we've done it in a couple of places, but we should actively avoid it if there's an alternative.

If the `state: gathered` is broken (has incorrect logic), then the entire premise of testing this module becomes invalid.

tests/integration/targets/nxos_vrf_interfaces/tests/common/gathered.yaml
roverflow: yes the upstream also has 64 interfaces, the above change only checks for those n interfaces that have a change in them

tests/integration/targets/nxos_vrf_interfaces/tests/common/_gather_interfaces.yaml
roverflow: i can do the same for asserting the before state as well, ill check if we are having a specified before state for those N interfaces we touch

NilashishC (approved):

PR #941, opened 2025-03-24
Fix lacp suspend individual command

msieland:
##### SUMMARY
Fixed the notation of the "lacp suspend individual command"

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
cisco.nxos.nxos_lacp_interfaces

##### ADDITIONAL INFORMATION
The notation is only wrong in the "del_attribs" function.
In the "add_commands" it is correct.
So it is only a problem if you try to change a lacp interface from "no lacp suspend-individual" to "lacp ssuspend-individual".


AAYUSH2091:
@msieland kindly add the changelogs and test to validate the changes.

msieland:
I added the changelogs and test. What are the next steps to implement the fix?

Issue #819, opened 2024-02-20
[lag_interfaces]


Egor-Ozhmegoff:
##### SUMMARY
I tried to create lag interfaces but i got an error.

FAILED! => {"changed": false, "module_stderr": "'NoneType' object is not iterable", "module_stdout": "", "msg": "MODULE FAILURE\nSee stdout/stderr for the exact error"}

I round the reason for this bug. That happened because i created port-channel without any child interfaces. I manually added one ethernet interface to port-channel and problem had gone.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
cisco.nxos.nxos_lag_interfaces

##### ANSIBLE VERSION
```paste below
ansible [core 2.16.2]
  config file = /home/eozhmegov/nautobot-device-config/ansible.cfg
  configured module search path = ['/home/eozhmegov/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/eozhmegov/.local/lib/python3.10/site-packages/ansible
  ansible collection location = /home/eozhmegov/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/eozhmegov/.local/bin/ansible
  python version = 3.10.6 (main, Aug  2 2022, 15:11:03) [GCC 7.5.0] (/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True
```

##### COLLECTION VERSION

```paste below
cisco.nxos  5.3.0
```

##### CONFIGURATION

```paste below

```

##### OS / ENVIRONMENT
Ubuntu 20.04

##### STEPS TO REPRODUCE

variables:
```yaml
lacp_interfaces:
  - group: 100
    ip: 192.168.0.0/31
    children:
      - Ethernet1/47
      - Ethernet1/48
```

role:
```yaml
---
- name: Interfaces base config
  cisco.nxos.nxos_interfaces:
    config: >-
      [
        {% for item in lacp_interfaces %}
        {
          "name": "port-channel{{ item.group }}",
          "description": "{{ item.description | default(omit)}}",
          "enabled": "{{ item.state | default(true) }}",
          "mtu": "{{ item.mtu | default("9216") }}",
          {% if item.ip is defined %}
          "mode": "layer3",
          {% else %}
          "mode": "layer2",
          {% endif %}
        },
        {% if item.children is defined %}
        {% for child in item.children %}
        {
          "name": "{{ child }}",
          "description": "{{ item.description | default(omit)}}",
[...]

PR #960, opened 2025-06-05
Issue #819: Fix replace LAG which has no existing members

ChrisPortman:
##### SUMMARY

Setting `state: replaced` on LAG interfaces when the existing config had no members resulted in an error:

*"MODULE FAILURE\nSee stdout/stderr for the exact error"*

This happened in the `intersect_list_of_dicts` which would receive `h=None` in this scenario.

This change adds checks to the `intersect_list_of_dicts` function in the same way per `diff_list_of_dicts`.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

cisco.nxos.nxos_lag_interfaces


ChrisPortman:
> @ChrisPortman Can you please add a tests to validate these changes and changelog as well.

Hi, I've added unit tests.  There were no existing unit tests for the `lag_interfaces` module, so I copied the LACP tests and modified.

PR #962, opened 2025-06-05
Added support for rd attribute for vrf_global module

nickbhasin:
SUMMARY
Added bug fix for 'rd' (Route Distinguisher) attribute addition for vrf_global module.

ISSUE TYPE
Bugfix Pull Request

COMPONENT NAME
cisco.nxos.nxos_vrf_global
INFO
This change covers the fix for 'rd' attribute support in vrf_global module with examples and tests.


KB-perByte:
@nickbhasin the changes look good, but the integration tests are failing.
Can you please check!

tests/integration/targets/nxos_vrf_global/tests/common/replaced.yaml
```diff
@@ -55,7 +56,7 @@
     - name: Assert that after dict is correctly generated
       ansible.builtin.assert:
         that:
-          - replaced['after'] == result['after']
+          - replaced['after'] | dict2items | difference(result['after'] | dict2items) | length == 0
```
Ruchip16, line 59: why are we doing dict2items?

tests/integration/targets/nxos_vrf_global/tests/common/overridden.yaml
```diff
@@ -64,7 +64,7 @@
     - name: Assert that before dict is correctly generated
       ansible.builtin.assert:
         that:
-          - "{{ overridden['after'] | symmetric_difference(result['before']) |length == 0 }}"
+          - "{{ overridden['after'] | dict2items | symmetric_difference(result['before'] | dict2items) | length == 0 }}"
```
Ruchip16, line 67: same here as well

Ruchip16 (changes requested):
i did not understand why we are doing dict2items here in certain test cases, rest changes looks fine

tests/integration/targets/nxos_vrf_global/tests/common/overridden.yaml
nickbhasin: It was failing due to order of keys was not same.

tests/integration/targets/nxos_vrf_global/tests/common/replaced.yaml
Ruchip16: we can use` sorted ` attribute then to mkae it in a certain order, wdyt?

nickbhasin: dict2items is a better choice since sorted only compares the keys and not the values. Also it will fail for nested items.

Issue #990, opened 2025-08-19
l3_interfaces ipv4.secondary: False is treated as True


ChrisPortman:
##### SUMMARY

given:

```yaml
config:
  - name: Ehternet1
    ipv4:
      - address: 10.1.1.1/24
        secondary: False
```

`secondary` will be added to the output command

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

module nxos_l3_interfaces

PR #991, opened 2025-08-19
Issue #990 Handle `secondary: false` for L3 interfaces Ipv4 address

ChrisPortman:
Correct issue where by false values for the setting were interpreted as true because the code only checked if the argument was defined.

This also now checks that it is True.

Fixes: #990

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

module nxos_l3_interfaces


plugins/module_utils/network/nxos/rm_templates/l3_interfaces.py
```diff
@@ -120,7 +120,7 @@ def __init__(self, lines=None, module=None):
             "setval": "ip address"
             "{{ ' ' + ipv4.address.address if ipv4.address.address is defined else '' }}"
             "{{ ' ' + ipv4.address.ip_network_mask if ipv4.address.ip_network_mask is defined else '' }}"
-            "{{ ' secondary' if ipv4.address.secondary is defined else '' }}"
+            "{{ ' secondary' if ipv4.address.secondary is defined and ipv4.address.secondary else '' }}"
```
nickbhasin, line 123: Please change it to "{{ ' secondary' if ipv4.address.secondary|d(False) else '' }}"

nickbhasin: Please add a changelog as well. Example - [](https://github.com/nickbhasin/cisco.nxos/blob/ee7969a8108e59e73df2d817f774b198ca61c776/changelogs/fragments/vrf_global_rd_attribute.yaml)

ChrisPortman: Changelog added

Issue #868, opened 2024-06-11
When removing Speed and Duplex order of operations yields error: ERROR: Duplex is not Compatible with Speed


bl1nkbl1nk:
##### SUMMARY

When speed and duplex settings are removed in nxos_interfaces the order of operations should be to delete duplex first and then speed. Otherwise NXOS gives an error.

```
router01(config)# interface ethernet 1/9
router01(config-if)# no speed
ERROR: Duplex is not Compatible with Speed

````

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
  cisco.nxos.nxos_interfaces:

##### ANSIBLE VERSION
```
ansible-prod % ansible --version
ansible [core 2.15.8]
```

##### COLLECTION VERSION

```
8.1.0
```

##### CONFIGURATION

```
no diff
```

##### OS / ENVIRONMENT

Ansible EE image

##### STEPS TO REPRODUCE

Try to run a config with speed and duplex to auto on an interface that currently is set to 100/full (manual)

```
  - description: some device
    duplex: full
    enabled: true
    mode: layer2
    name: Ethernet1/5
    speed: '100'
    
 Then after that configure this with duplex and speed removed (default auto)
 
   - description: some device
    enabled: true
    mode: layer2
    name: Ethernet1/5

 
```

##### EXPECTED RESULTS

when casting "speed" and "duplex" 

When configuring 
1. Speed
2. Duplex

When removing (no speed)
1. Duplex
2. Speed

##### ACTUAL RESULTS

Module tries to delete speed first and gets error from device:

ERROR: Duplex is not Compatible with Speed

##### Possible solution??

in line 380 change the order of operations? I'm not super aware of how this codebase works, but looks like the right area

https://github.com/ansible-collections/cisco.nxos/blob/19b2473a1ab1638a18eed32dd08dd2662853610f/plugins/module_utils/network/nxos/config/interfaces/interfaces.py#L380

PR #1052, opened 2026-03-16
Refactor interface parser for duplex and speed command handling

sunshinegorgon7:
Refactor interface parser to handle duplex and speed

## Description
- **What is being changed?**
  `duplex` and `speed` have been removed from the `parsers` list in the `Interfaces` resource module and are now handled explicitly in the `_compare` method with duplex always generated before speed.

- **Why is this change needed?**
  NX-OS requires `duplex` to be configured before `speed` when transitioning to or from `auto`. When both were in the `parsers` list, `self.compare()` iterated them in declaration order (speed first, duplex second), causing NX-OS to return:
  `ERROR: Duplex is not Compatible with Speed`
  This made it impossible to set `speed auto` / `duplex auto` via the module in a single push, forcing users to either manually sequence two separate plays or bypass the module entirely with raw CLI — neither of which is suitable for CI/CD pipelines.

- **How does this change address the issue?**
  By handling `duplex` and `speed` explicitly in `_compare`, we control the exact command generation order. Idempotency is preserved — commands are only generated when the desired value differs from the current state. Additionally, when `auto` is desired and the device has no explicit speed/duplex config (NX-OS default), no commands are generated to prevent spurious `changed` on subsequent runs.

## Type of Change
- [x] Refactoring (no functional changes)

## Related Issue
- Related to #868

## Component Name
`nxos_interfaces`

## Self-Review Checklist

### Prerequisites
- NX-OS Version: 10.3(8)
- Hardware Platform: Cisco Nexus 9000
- Test Topology: Single NX-OS device with at least one physical Ethernet interface available for configuration

### Steps to Test

**Scenario 1: Set explicit speed and duplex**
1. Define an interface var with `speed: "1000"` and `duplex: full`
2. Run `nxos_interfaces` with `state: merged`
3. Verify interface reflects `speed 1000` and `duplex full` on device

**Scenario 2: Transition from explicit to auto (the bug scenario)**
1. Starting from Scenario 1 state, update the var to `speed: "auto"` and `duplex: auto`
2. Run `nxos_interfaces` with `state: merged` in a single push
3. Verify no `ERROR: Duplex is not Compatible with Speed` error
4. Verify interface reflects `speed auto` and `duplex auto` on device

**Scenario 3: Idempotency check**
1. Run Scenario 2 a second time without any changes
2. Verify task returns `ok` with no changes applied

> Note: When `speed: auto` / `duplex: auto` are set, NX-OS does not write
> explicit config to the running config — the interface returns to its default
> state with no speed/duplex lines. The fix accounts for this by treating
> `auto` in `want` with no corresponding `have` value as already at the
> desired default state, preventing spurious `changed` on subsequent runs.

### Expected Results
- All three scenarios complete without errors
- Scenario 2 succeeds in a single task execution without manual sequencing
- Scenario 3 returns `ok`

### Test Results

```
========
main.yml
========
- name: NXOS Interfaces
  hosts: reachable
  gather_facts: false
  vars:
    ansible_connection: ansible.netcommon.network_cli
    ansible_network_os: cisco.nxos.nxos
    ansible_user: ansible
    ansible_ssh_private_key_file: "{{ KEY }}"
  tasks:
    - name: Include NX-OS interface role
      ansible.builtin.include_role:
        name: nxos_interface
      when: hostvars[inventory_hostname].platforms[0] == 'nxos'

===================================
roles/nxos_interface/tasks/main.yml
===================================
- name: Prepare list of Nexus interface configurations
  ansible.builtin.set_fact:
    prep_nexus_int: "{{ prep_nexus_int | default([]) +
    [{'name': int.name, 'description': (int.description | default(omit)),
    'speed': (int.speed | default(omit)), 'duplex': (int.duplex | default(omit)),
    'enabled': (int.enabled | default(omit)), 'mode': (int.mode | default(omit)),
    'mtu': (int.mtu | default(omit))}] }}"
  loop: "{{ interfaces }}"
  loop_control:
    loop_var: int

- name: Push Nexus Interface settings
  cisco.nxos.nxos_interfaces:
    config: "{{ prep_nexus_int }}"
    state: merged

===================
hostvars/POC-N9K-01
===================
interfaces:
  - name: Ethernet1/5
    description: '** PO TEST **'
    enabled: true
    speed: auto
    duplex: auto
    mode: layer2
```

## Acceptance Criteria Verification
- [x] All acceptance criteria have been reviewed and verified
- [x] Acceptance criteria checklist:
  - [x] `duplex auto` + `speed auto` can be set in a single merged push
  - [x] Explicit speed/duplex values (e.g. 1000/full) still work correctly
  - [x] Idempotency is maintained — no changes when desired state matches current state
  - [x] No regression on other parsers (description, mtu, mode, enabled, etc.)

## Additional Context

The root cause is in `interfaces.py`. The original `parsers` list passed both `speed` and `duplex` to `self.compare()`, which generates commands in list order — speed before duplex. NX-OS enforces that duplex must be compatible with the current speed before it can be changed, which means when clearing back to `auto`, duplex must be set first.

**Before (broken for auto transition):**

```python
self.parsers = [
    "description",
    "speed",    # generates speed command first
    "mtu",
    "duplex",   # generates duplex command second — NX-OS rejects this order
    ...
]
```

**After (correct ordering):**

```python
# speed and duplex removed from parsers list
# handled explicitly in _compare():

# duplex always before speed
if want_duplex is not None and want_duplex != have_duplex:
    # if want is auto and have has no config, device is already at default — skip
    if not (want_duplex == "auto" and have_duplex is None):
        self.addcmd(want, "duplex", False)

if want_speed is not None and want_speed != have_speed:
    if not (want_speed == "auto" and have_speed is None):
        self.addcmd(want, "speed", False)
```

### Command Output / Logs

```
# Before fix:
FAILED: "speed auto\r\r\nERROR: Duplex is not Compatible with Speed"

# After fix:
changed: [POC-N9K-01] => interface Ethernet1/5: duplex auto, speed auto applied successfully
```

### Required Actions
- [x] Requires changelog fragment
  - Added `changelogs/fragments/nxos_interfaces_fix_duplex_speed_ordering.yaml`
- [x] Requires unit test updates
  - Added `test_duplex_before_speed_ordering` — asserts duplex command index is lower than speed command index in raw (unsorted) command output
  - Added `test_auto_speed_duplex_idempotency` — asserts no commands generated when `auto` is desired and device has no existing speed/duplex config
- [ ] Requires integration test updates


sunshinegorgon7:
Thoughts?

PR #1069, opened 2026-05-29
Update bgp_global to add template prefix to template sections coming from device config

TheNova22:
## Description

- What is being changed?
nxos_bgp_global is currently also reading template blocks within the query(`show running-config | section '^router bgp'`) which is not required as bgp_template module exists for that activity.
- Why is this change needed?
Currently if any bgp config contains template blocks, it is read and can cause issues like the following:
```
router bgp 65137
  template peer TC10-TEMPLATE-NXOS-1
    local-as 65200
```
```
fatal: [nxos-jay]: FAILED! => {"changed": false, "msg": "Unsupported parameters for (basic.py) module: config.local_as. Supported parameters include: affinity_group, as_number, bestpath, cluster_id, confederation, disable_policy_batching, dynamic_med_interval, enforce_first_as, enhanced_error, fabric_soo, fast_external_fallover, flush_routes, graceful_restart, graceful_shutdown, isolate, log_neighbor_changes, maxas_limit, neighbor_down, neighbors, nexthop, rd, reconnect_interval, router_id, shutdown, suppress_fib_pending, timers, vrfs."}
```
Reason: It detected the local-as line as a vrf sub-item, but no vrf exists as parent here, so it breaks. [Reference Line](https://github.com/ansible-collections/cisco.nxos/blob/main/plugins/module_utils/network/nxos/rm_templates/bgp_global.py#L1537).
- How does this change address the issue?
It ignore all template references within the query so that change happens only at the bgp global level.

## Type of Change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] New module (adding a new module to the collection)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Test update
- [ ] Refactoring (no functional changes)
- [ ] Collection release
- [ ] CI maintenance
- [ ] Workflow maintenance
- [ ] Configuration change

## Component Name

nxos_bgp_global

## Self-Review Checklist

### DIFF
--- a/plugins/module_utils/network/nxos/config/bgp_global/bgp_global.py
+++ b/plugins/module_utils/network/nxos/config/bgp_global/bgp_global.py
@@ -367,6 +367,7 @@ class Bgp_global(ResourceModule):
         vrf_data = {}
 
         for x in data:
+            # No need to prepend template lines as each line is checked for keyword highlighting
             if x.strip().startswith("vrf"):
                 cur_nbr = None
                 cur_vrf = x.split(" ")[-1]
--- a/plugins/module_utils/network/nxos/config/bgp_global/bgp_global.py
+++ b/plugins/module_utils/network/nxos/config/bgp_global/bgp_global.py
@@ -363,10 +363,26 @@ class Bgp_global(ResourceModule):
         data = self._get_config().split("\n")
         cur_nbr = None
         cur_vrf = None
+        in_template_cxt = False
+        cur_template = {}
         gbl_data = set()
         vrf_data = {}
 
         for x in data:
+            cur_indent = len(x) - len(x.lstrip())
+
+            # Skip template peer* blocks - they are managed by nxos_bgp_templates
+            if x.strip().startswith("template peer"):
+                in_template_cxt = True
+                cur_template["indent"] = cur_indent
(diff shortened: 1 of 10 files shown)


TheNova22:
configs in the unit and integration tests were updated with the expectation that their output/generated commands should not change.

plugins/module_utils/network/nxos/config/bgp_global/bgp_global.py
```diff
@@ -363,10 +363,26 @@ def _build_af_data(self):
         data = self._get_config().split("\n")
         cur_nbr = None
         cur_vrf = None
+        in_template_cxt = False
```
rohitthakur2590: let's reevaluate the need to reconnect to the device here and refetching the data, IMO we should be relying on the already parsed data.

rohitthakur2590 (approved):

plugins/module_utils/network/nxos/config/bgp_global/bgp_global.py
TheNova22: Yup, I'll have a separate ticket created for this as it is independent of this change.

PR #1075, opened 2026-06-23
HSRP Interfaces | Update preempt template and fix replaced state behaviour for it

TheNova22:
## Description
  

  **What is being changed?**

  This PR fixes two bugs in the `cisco.nxos.nxos_hsrp_interfaces` resource module affecting `preempt` delay sub-option handling, and a `priority forwarding-threshold` template
  defect. It also introduces targeted integration and unit tests to cover the repaired code paths.

  **Why is this change needed?**

  NX-OS HSRP `preempt delay` is configured as a single compound CLI command (`preempt delay minimum <N> reload <N> sync <N>`), but each sub-key (`minimum`, `reload`, `sync`) can be removed individually with `no
  preempt delay <sub-key> <value>`. The module previously treated the entire `preempt` block as a monolithic scalar value (like `group_name` or `mac_address`), so:

  1. When `state: replaced` or `overridden` was invoked with only a subset of preempt sub-keys, the module compared the whole dict and — because the *present* sub-keys matched — generated no commands at all,
  leaving unwanted sub-keys (e.g., `reload`) intact on the device.
  2. The `preempt` setval template unconditionally emitted `preempt delay minimum ...` as its first word, even when `minimum` was absent, producing malformed commands like `preempt delay minimum None reload 50`.
  3. The `priority` setval template emitted `forwarding-threshold lower <N> upper <N>` without the `forwarding-threshold` keyword when `upper` was set without `lower`.

  **How does this change address the issue?**

  - **Architecture split** — `complex_parsers` is divided into `complex_scalar_parsers` (single-value parsers that use the existing compare/negate path) and `complex_dict_parsers` (dict-valued parsers that
  require per-sub-key handling).
  - **New `_complex_compare` method** — for `replaced`/`overridden` states, iterates sub-keys present in `have` but absent in `want` and emits an individual `no preempt delay <sub-key>` command for each. Only
  then emits the positive `preempt delay ...` command if any desired sub-keys changed.
  - **`preempt` setval template fix** — `minimum` is now conditional like `reload` and `sync`, so the command is built only from the sub-keys actually specified.
  - **`priority` setval template fix** — `forwarding-threshold` keyword is guarded by `if lower is defined or upper is defined`, ensuring correct CLI when either bound is present.

  **Concrete reproduction of the primary bug:**

  Device state before the playbook runs:
  ```yaml
  - name: Vlan218
    standby_options:
      - group_no: 218
        preempt:
          reload: 20      # <<< exists on device
        priority:
          level: 110
```
  Playbook invocation (state: replaced — intent: keep only sync: 55):
```
  cisco.nxos.nxos_hsrp_interfaces:
    config:
      - name: "Vlan218"
        standby_options:
          - group_no: 218
            preempt:
              sync: 55
            priority:
              level: 55
    state: replaced
```
  Before this fix:
```
  result:
    changed: false
    commands: []     # BUG: expected 'no preempt delay reload 20'
```
  After this fix:
```
  result:
    changed: true
    commands:
      - interface Vlan218
      - hsrp 218
      - no preempt delay reload 20
      - preempt delay sync 55
```
 ### Type of Change

  
  - [ ] New feature (non-breaking change which adds functionality)
  - [ ] New module (adding a new module to the collection)
  - [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
  - [ ] Documentation update
  - [x] Test update
  - [ ] Refactoring (no functional changes)
  - [ ] Collection release
  - [ ] CI maintenance
  - [ ] Workflow maintenance
  - [ ] Configuration change

  ### Component Name

  
  nxos_hsrp_interfaces

 ###  Self-Review Checklist

  
Assisted By: Claude Code

### DIFF
--- a/changelogs/fragments/hsrp_interfaces_partial_dict_options.yaml
+++ b/changelogs/fragments/hsrp_interfaces_partial_dict_options.yaml
@@ -1,0 +1,11 @@
+---
+minor_changes:
+  - nxos_hsrp_interfaces - In ``replaced`` and ``overridden`` states, negate removed
+    sub-options for ``preempt`` before applying the desired configuration. This aligns
+    command generation with NX-OS additive CLI behavior when only part of a preempt
+    delay setting is specified.
+bugfixes:
+  - nxos_hsrp_interfaces - Update ``preempt`` and ``priority`` templates to render
+    optional sub-options independently, enabling per-key command generation.
+  - nxos_hsrp_interfaces - Normalize gathered timer values before compare to improve
+    idempotency for second-based ``hello_interval`` and ``hold_time`` settings.
--- a/changelogs/fragments/hsrp_interfaces_partial_dict_options.yaml
+++ b/changelogs/fragments/hsrp_interfaces_partial_dict_options.yaml
@@ -7,5 +7,3 @@ minor_changes:
 bugfixes:
   - nxos_hsrp_interfaces - Update ``preempt`` and ``priority`` templates to render
     optional sub-options independently, enabling per-key command generation.
-  - nxos_hsrp_interfaces - Normalize gathered timer values before compare to improve
-    idempotency for second-based ``hello_interval`` and ``hold_time`` settings.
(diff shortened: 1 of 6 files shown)


rohitthakur2590 (approved):
LGTM!

PR #1089, opened 2026-08-04
Fix nxos_route_maps fact gather crash on match metric +- deviation

malikdotexe:
## Summary

Fixes fact gathering / parsed-state crashes in `nxos_route_maps` (and `nxos_facts` with `gather_network_resources: [route_maps]`) when a route-map entry uses NX-OS **metric deviation** syntax:

```
match metric <value> +- <deviation>
```

Resolves Red Hat support case [04500155](https://access.redhat.com/support/cases/#/case/04500155).

> Note: The failure came from route-map that used `match metric 0 +- 100`.

## What was going wrong

`match.metric` was defined as:

```python
"metric": {"type": "list", "elements": "int"}
```

The parser captured everything after `match metric` and split on whitespace:

```
match metric 0 +- 100
→ ["0", "+-", "100"]
```

`validate_config` then tried to coerce `"+-"` to `int` and aborted:

```
Elements value for option 'metric' found in 'config -> entries -> match'
is of type <class 'str'> and we were unable to convert to int:
<class 'str'> cannot be converted to an int
```

This is valid NX-OS CLI (confirmed on Nexus / NX-OS 10.4(2)). Multiple value+deviation pairs can also appear on one line:

```
match metric 0 +- 100 100 +- 65000
```

## What we changed

Modeled `match.metric` as a list of dicts (same idea as IOS route-map metric deviation support, adapted for NX-OS multi-pair lines):

| CLI | Parsed fact |
| --- | --- |
| `match metric 100 200` | `[{"value": 100}, {"value": 200}]` |
| `match metric 0 +- 100` | `[{"value": 0, "deviation": 100}]` |
| `match metric 0 +- 100 100 +- 65000` | `[{"value": 0, "deviation": 100}, {"value": 100, "deviation": 65000}]` |

Changes:

1. **Argspec / module docs** — `match.metric` is now `list` of dicts with required `value` (`int`) and optional `deviation` (`int`).
2. **rm_templates** — custom `setval` renders `+-` when present; `parse()` normalizes split tokens into dicts before validation.
3. **config compare** — `_compare_lists` handles list-of-dicts for `match.metric` (no longer assumes hashable ints).
4. **Unit tests** — parsed + merged coverage for deviation; existing plain-metric merged test updated to the new shape.
5. **Integration test** — `tests/integration/targets/nxos_route_maps/tests/common/match_metric_deviation.yaml` covers:
   - seed CLI `match metric +-` then `state: gathered`
   - `nxos_facts` with `gather_network_resources: [route_maps]`
   - `state: merged` command generation + idempotency
6. **Changelog fragment** — bugfix entry.

### Playbook shape (after this change)

```yaml
match:
  metric:
    - value: 0
      deviation: 100
    - value: 100
      deviation: 65000
```

Plain metrics (no deviation):

```yaml
match:
  metric:
    - value: 100
    - value: 200
```

> **Note:** this changes the `match.metric` argspec from `list[int]` to `list[dict]`. Existing playbooks that used `metric: [100, 200]` need to switch to the dict form above.

## How this solves it

Tokens like `+-` are no longer treated as metric element values. Each CLI metric/deviation pair maps to one dict, so `validate_config` succeeds and fact gathering / `state: parsed` work for real customer configs that mix as-path, prefix-lists, and metric deviation (as in case 04500155).

## Reproduction results (before / after)

Live lab repro on NX-OS 10.4(2) with:

```
route-map ANSIBLE-METRIC-DEV-TEST permit 10
  match metric 0 +- 100 100 +- 65000
route-map EXAMPLE permit 10
  match as-path SOME-NAMED-SET
```

### Before

```
TASK [Gather route_maps network resource facts]
fatal: [nxos]: FAILED! => {
  "msg": "Elements value for option 'metric' found in 'config -> entries -> match' is of type <class 'str'> and we were unable to convert to int: <class 'str'> cannot be converted to an int"
}
```

### After (this PR)

`cisco.nxos.nxos_facts` with `gather_network_resources: [route_maps]` succeeds:

```json
{
  "route_maps": [
    {
      "route_map": "ANSIBLE-METRIC-DEV-TEST",
      "entries": [
        {
          "sequence": 10,
          "action": "permit",
          "match": {
            "metric": [
              {"value": 0, "deviation": 100},
              {"value": 100, "deviation": 65000}
            ]
          }
        }
      ]
    },
    {
      "route_map": "EXAMPLE",
      "entries": [
        {
          "sequence": 10,
          "action": "permit",
          "match": {
            "as_path": ["SOME-NAMED-SET"]
          }
        }
      ]
    }
  ]
}
```

Customer-shaped config from case 04500155 (excerpt) also validates after the fix, e.g.:

```
route-map ARY-FROM-MSO-BRC permit 10
  match ip address prefix-list FWG-NAT-POOLS-V4
  match metric 0 +- 100
  set community 65139:147
```

## Test plan

- [x] Unit: `test_nxos_route_maps_match_metric_deviation_parsed`
- [x] Unit: `test_nxos_route_maps_match_metric_deviation_merged`
- [x] Unit: full `test_nxos_route_maps.py` (23 passed)
- [x] Integration: `match_metric_deviation.yaml` on live NX-OS (gathered + `nxos_facts` + merged idempotent)
- [x] Live NX-OS repro: push `match metric 0 +- 100` / multi-pair line, gather via `nxos_facts` (before crash / after success)
- [x] Confirm `match as-path <named-acl>` continues to parse as `as_path` (list of str) and is not involved in the failure
- [ ] CI unit / sanity on PR

### DIFF
--- a/changelogs/fragments/route_maps_match_metric_deviation.yaml
+++ b/changelogs/fragments/route_maps_match_metric_deviation.yaml
@@ -1,10 +1,7 @@
 ---
-bugfixes:
-  - nxos_route_maps - Fix fact gathering / parsed-state crash when route-map
-    entries use NX-OS metric deviation syntax
-    (``match metric <value> +- <deviation>``). Previously the parser split
-    tokens such as ``0 +- 100`` into ``["0", "+-", "100"]`` and
-    ``validate_config`` failed because ``match.metric`` was typed as a list
-    of integers. ``match.metric`` elements are now dictionaries with required
-    ``value`` (replacing the previous list-of-int form) and a new optional
-    ``deviation`` parameter (support case 04500155).
+minor_changes:
+  - nxos_route_maps - Add support for NX-OS metric deviation syntax
+    (``match metric <value> +- <deviation>``) and fix fact gathering /
+    parsed-state crash on that CLI. ``match.metric`` elements are now
+    dictionaries with required ``value`` (replacing the previous list-of-int
+    form) and a new optional ``deviation`` parameter (support case 04500155).
--- a/changelogs/fragments/route_maps_match_metric_deviation.yaml
+++ b/changelogs/fragments/route_maps_match_metric_deviation.yaml
@@ -5,5 +5,6 @@ bugfixes:
     (``match metric <value> +- <deviation>``). Previously the parser split
     tokens such as ``0 +- 100`` into ``["0", "+-", "100"]`` and
     ``validate_config`` failed because ``match.metric`` was typed as a list
-    of integers. ``match.metric`` now models a list of dicts with
-    ``value`` and optional ``deviation`` (support case 04500155).
+    of integers. ``match.metric`` elements are now dictionaries with required
+    ``value`` (replacing the previous list-of-int form) and a new optional
+    ``deviation`` parameter (support case 04500155).
--- a/changelogs/fragments/route_maps_match_metric_deviation.yaml
+++ b/changelogs/fragments/route_maps_match_metric_deviation.yaml
@@ -1,0 +1,9 @@
+---
+bugfixes:
+  - nxos_route_maps - Fix fact gathering / parsed-state crash when route-map
+    entries use NX-OS metric deviation syntax
+    (``match metric <value> +- <deviation>``). Previously the parser split
+    tokens such as ``0 +- 100`` into ``["0", "+-", "100"]`` and
+    ``validate_config`` failed because ``match.metric`` was typed as a list
+    of integers. ``match.metric`` now models a list of dicts with
+    ``value`` and optional ``deviation`` (support case 04500155).
--- a/changelogs/fragments/route_maps_match_metric_deviation.yaml
+++ b/changelogs/fragments/route_maps_match_metric_deviation.yaml
@@ -1,7 +1,8 @@
 ---
-minor_changes:
-  - nxos_route_maps - Add support for NX-OS metric deviation syntax
-    (``match metric <value> +- <deviation>``) and fix fact gathering /
-    parsed-state crash on that CLI. ``match.metric`` elements are now
-    dictionaries with required ``value`` (replacing the previous list-of-int
-    form) and a new optional ``deviation`` parameter (support case 04500155).
+breaking_changes:
+  - >-
+    nxos_route_maps - ``match.metric`` elements changed from a list of integers
+    to a list of dictionaries with required ``value`` and optional ``deviation``
+    (for NX-OS ``match metric <value> +- <deviation>``). Update playbooks from
+    ``metric: [10, 200]`` to ``metric: [{value: 10}, {value: 200}]`` (or add
+    ``deviation`` when needed). Fixes fact-gathering failures on that CLI.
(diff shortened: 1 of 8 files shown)


changelogs/fragments/route_maps_match_metric_deviation.yaml
```diff
@@ -0,0 +1,9 @@
+---
+bugfixes:
```
TheNova22: Lets update the changelog to ensure we mention that a new parameter has been added called `deviation` and the `metric` value type is now `dict` instead of `list`.

TheNova22 (changes requested):

changelogs/fragments/route_maps_match_metric_deviation.yaml
malikdotexe: Updated — changelog now explicitly calls out that `match.metric` elements are dictionaries with required `value` (replacing the previous list-of-int form) and the new optional `deviation` parameter. Also addressed the Sonar S1192 finding by extracting a `MATCH_METRIC` constant.

plugins/module_utils/network/nxos/config/route_maps/route_maps.py
```diff
@@ -93,7 +96,7 @@ def __init__(self, module):
             "match.ipv6.next_hop.prefix_lists",
             "match.ipv6.route_source.prefix_lists",
             "match.mac_list",
-            "match.metric",
+            MATCH_METRIC,
```
TheNova22: I don't think we should deviate away from string based parser key reference. 

If sonar is asking you to change this. Revert and then you can go to sonar dashboard and mark it as invalid error(or something along the lines of that).

TheNova22 (changes requested):

changelogs/fragments/route_maps_match_metric_deviation.yaml
```diff
@@ -0,0 +1,10 @@
...
```
komaldesai13: As we are introducing new sub options , we would consider it as minor change

plugins/module_utils/network/nxos/config/route_maps/route_maps.py
komaldesai13: +1

komaldesai13 (commented):
rest looks good

komaldesai13 (changes requested):

changelogs/fragments/route_maps_match_metric_deviation.yaml
malikdotexe: Updated — moved this to `minor_changes` since we are introducing the new `deviation` suboption (and changing `match.metric` element type to dict).

plugins/module_utils/network/nxos/config/route_maps/route_maps.py
malikdotexe: Agreed , I have reverted to string-based `"match.metric"` parser key references. Will mark the Sonar S1192 finding as Won't Fix / False Positive on the dashboard so we keep the conventional string keys.

komaldesai13 (approved):

TheNova22 (approved):

PR #1092, opened 2026-08-17
Fix nxos_snmp_server community removal on replaced/overridden states

malikdotexe:
## Summary

Fixes a bug where `cisco.nxos.nxos_snmp_server` with `state: replaced` (or `overridden`) **failed to remove SNMP v2c communities**. The module correctly decided to delete extra communities, but the generated CLI immediately re-created them in the same run. That left unwanted communities on the device and caused nightly policy-enforcement jobs to report `changed: true` on every run.

This PR also **re-enables `network_cli` and NX-API (`httpapi`) integration tests** for `nxos_snmp_server` and fixes the community round-trip bugs that made those tests fail, so CI can catch this class of issue.

Related customer ticket: [AAP-87440](https://redhat.atlassian.net/browse/AAP-87440)

---

### What was going wrong

When a community is removed, `_compare_lists()` calls `addcmd(..., negate=True)`. Because the `communities` parser had **no `remval`**, netcommon fell back to negating the **multi-line `setval`** template used to create communities:

```
snmp-server community {{ name }} {{ group }}
snmp-server community {{ name }} {{ ro/rw }}
snmp-server community {{ name }} {{ acls }}
```

`NetworkTemplate._render()` only prefixes `no ` on the **entire string**, not each line. The switch therefore received:

```text
no snmp-server community private  group network-operator
snmp-server community private
snmp-server community private
```

Line 1 deletes the community; lines 2–3 add it back.

### Fix

Add a dedicated `remval` for communities:

```python
"remval": "snmp-server community {{ name }}",
```

Deletes now emit a single command:

```text
no snmp-server community private
```

---

## Part 2 — Why integration tests were off, and how we turned them back on

SNMP CLI/NX-API includes in `tasks/main.yaml` were commented with `# uncomment later` in [#1014](https://github.com/ansible-collections/cisco.nxos/pull/1014). Enablement never happened, so CI never ran `replaced`/`merged`/`overridden`. That is why AAP-87440 was not caught.

Uncommenting CLI was not enough. Two more bugs still broke **merged idempotency**:

### NX-OS stores one community as several lines

```
snmp-server community private group network-admin
snmp-server community private use-ipv4acl MY_ACL
```

The parser treated each line as a **different** list entry. Want was one object (`name` + `group` + ACL). Have was two objects. Compare never matched, so the second `merged` run always changed config.

**Fix:** after parse, merge all lines with the same community name into one dict. Also parse `ro`/`rw` when those keywords are present.

### Empty `setval` lines reset access to read-only

`setval` always rendered extra bare `snmp-server community NAME` lines. On NX-OS that resets the community to default read-only (`group network-operator`). We wrote `network-admin`; the device stored `network-operator`; the next merge ran again.

**Fix:** emit only the community lines for attributes that are actually set. The merged test no longer sets `ro: true` (NX-OS does not persist `ro` in running-config on the lab image, and it fights with an explicit `network-admin` group).

---

## Part 3 — NX-API integration enablement

`nxapi.yaml` is now uncommented alongside `cli.yaml`. Enabling NX-API required two additional changes because **NX-API fails the entire edit on any CLI error**, whereas `network_cli` often continues after a bad line.

### Community commands must be one line per attribute

Multi-line community `setval` strings (newline-joined) work over `network_cli` but NX-API rejects them as `% Invalid command`.

**Fix:** `_template_communities()` returns a list of single-line commands (group, ro/rw, ACLs as separate entries). `addcmd`/`to_list` already flatten lists into the command batch.

### Test cleanup cannot use `state: deleted`

`_remove_config.yaml` previously gathered config and applied `state: deleted`. That fails over NX-API because:

- `no snmp-server user admin` is rejected (`admin account cannot be deactivated`)
- Two `no snmp-server host` lines for the same IP fail when the first delete already removed the host target

**Fix:** capture baseline SNMP config at the start of the role (`state: gathered`) and restore it with `state: overridden`. Over NX-API, restore commands are applied one at a time via `nxos_config` with `ignore_errors: true` so benign "already gone" host deletes do not fail the play.

---

## Part 4 — CI failure: `ansible account cannot be deactivated`

After tests were re-enabled, CML integration failed on `replaced`/`overridden` with:

```text
CLI execution error: ansible account cannot be deactivated
```

### Why it passed locally but failed in CI

`replaced`/`overridden` delete any SNMP user that is on the device but not in the playbook list. NX-OS refuses `no snmp-server user <login>` for the account currently used to connect.

| Lab | Login user | Included in the original test `want` list? | Result |
|-----|------------|--------------------------------------------|--------|
| Local (`inventory.ini`) | `cisco` | yes (hardcoded with `admin`) | login user not deleted → pass |
| CML CI | `ansible` | no | module emitted `no snmp-server user ansible` → fail |

This is a **test** issue, not a module bug. The module is doing what `replaced`/`overridden` means.

### Fix

Do not hardcode `admin`/`cisco`. Gather baseline SNMP users at role start (already captured as `snmp_baseline_config`) and merge them into the `replaced`/`overridden` payload together with the test user (`snmp_user_1`). Command asserts already ignore `snmp-server user` lines because default users vary by image.

Verified locally: `replaced` + `overridden` pass on both `network_cli` and `httpapi`.

---

## Test plan

- [x] `ansible-test sanity` on changed files
- [x] Unit tests: `test_nxos_snmp_server.py` — **24 passed**
- [x] Live `ansible-test network-integration --allow-destructive nxos_snmp_server` (`network_cli`, NX-OS 10.4 lab) — **96 ok, 0 failed** (5 ignored on purpose in `empty_config`)
- [x] Live `ansible-test network-integration --allow-destructive nxos_snmp_server` (`httpapi`, same lab) — **100 ok, 0 failed** (8 ignored on purpose in `empty_config`)
- [x] Re-run `replaced` + `overridden` after baseline-user merge (`network_cli` and `httpapi`) — **0 failed**
- [ ] CI network_cli / NX-API jobs for `nxos_snmp_server` (re-run after this commit)

### Local integration recap

After `remval` + community merge + selective `setval` + NX-API fixes + baseline SNMP user merge:

- `replaced` / `overridden`: extra community removed with a single `no snmp-server community <name>`; login SNMP user is preserved; second run idempotent
- `merged`: second run `changed: false`
- `rendered` / `gathered` / `parsed` / `empty_config`: pass on both `network_cli` and `httpapi`


tests/integration/targets/nxos_snmp_server/vars/main.yml
TheNova22, line 1: Can we add the new entries as a list instead of using `"` and `\n` (if needed) to make it consistent with other command entries

TheNova22 (changes requested):

tests/integration/targets/nxos_snmp_server/vars/main.yml
malikdotexe: Addressed in the new commit

tests/integration/targets/nxos_snmp_server/tests/cli/.gitkeep
komaldesai13, line 1: why this needed

tests/integration/targets/nxos_snmp_server/tests/common/_remove_config.yaml
komaldesai13, line 1: we are not testing "`Deleted`" state here ?

tests/integration/targets/nxos_snmp_server/tests/common/_assert_commands.yaml
komaldesai13, line 1: why we need this file here

komaldesai13 (changes requested):

tests/integration/targets/nxos_snmp_server/tests/cli/.gitkeep
malikdotexe: Removed in the latest commit, empty tests/cli / tests/nxapi placeholders are not used elsewhere in this collection.

tests/integration/targets/nxos_snmp_server/tests/common/_remove_config.yaml
malikdotexe: Agreed , _remove_config is cleanup only; tests/common/deleted.yaml now covers state: deleted (login SNMP user is still not removed, NX-OS will not deactivate it).

tests/integration/targets/nxos_snmp_server/tests/common/_assert_commands.yaml
malikdotexe: Removed, command asserts are inlined with symmetric_difference, same pattern as the other resource modules.

komaldesai13 (approved):

TheNova22 (approved):

PR #1093, opened 2026-08-19
Re-enable and stabilize nxos_snmp_server network_cli integration tests

malikdotexe:
## Summary

This PR turns `nxos_snmp_server` **network_cli integration tests back on**, and fixes the module bugs that made those tests fail.

**Depends on / includes [#1092](https://github.com/ansible-collections/cisco.nxos/pull/1092)** (community `remval` so `replaced`/`overridden` actually delete communities). The new work for this PR is commit `60473e6b`.

`nxapi` is still left commented. That can be a follow-up once CLI is green in CI.

---

## Why were these tests off?

In [#1014](https://github.com/ansible-collections/cisco.nxos/pull/1014) the SNMP CLI/NX-API includes in `tasks/main.yaml` were commented with `# uncomment later`. Enablement never happened.

That meant CI never ran `merged` / `replaced` / `overridden` against a real device. Customer bug AAP-87440 (communities not removed on `replaced`) slipped through. After #1092 we uncommented CLI locally and the suite still failed — so this PR fixes the remaining bugs and turns CLI tests back on.

---

## What NX-OS actually does 

One SNMP community is **one name**, but the switch stores it as **several CLI lines**:

```
snmp-server community private group network-admin
snmp-server community private use-ipv4acl MY_ACL
```

Those two lines are still the same community (`private`). One line is the group, one line is the ACL.

`ro` / `rw` work the same way (`snmp-server community private ro`). On the lab NX-OS 10.4 image, `ro` is **not** kept in running-config. Default read-only access shows up as `group network-operator`.

---

### What happened

The parser matched **one running-config line at a time** and appended a new list entry each time. Gather then looked like:

```yaml
communities:
  - {name: private, group: network-admin}
  - {name: private, use_ipv4acl: MY_ACL}
```

The playbook asked for **one** object:

```yaml
- name: private
  group: network-admin
  use_ipv4acl: MY_ACL
```

Compare uses the **whole dict** as the key, so want and have never matched. Every `merged` rerun generated community commands again → **idempotency failed**.

The parser also ignored `ro` / `rw` even when those keywords were present.

### What we did

After parse, **merge all lines with the same community name into one dict**. Gather now returns:

```yaml
- name: private
  group: network-admin
  use_ipv4acl: MY_ACL
```

We also parse `ro` / `rw` when they appear in config.

This is in `facts/snmp_server.py` (`_merge_communities`) and the communities `getval`/`result` in `rm_templates/snmp_server.py`.

---

### What happened

Community `setval` **always** rendered three lines, even when the user only set a group:

```
snmp-server community private group network-admin
snmp-server community private          ← empty extra command
snmp-server community private          ← another empty extra command
```

On NX-OS, a bare `snmp-server community NAME` resets that community to **default read-only**. Running-config then showed `group network-operator` instead of `network-admin`.

So we wrote `network-admin`, the device stored `network-operator`, and the next merge thought it still needed to change. That is the rest of the old “merged idempotency / ro round-trip” failure.

`ro: true` in the merged test made this worse: it fights with `group: network-admin`, and `ro` is not stored in running-config on this platform.

### What we did

`setval` now emits **only the lines for attributes that are actually set**:

- group only → `snmp-server community private group network-admin`
- group + ACL → those two lines, nothing else
- no empty `snmp-server community NAME`

Deletes are unchanged: `remval` is still `no snmp-server community {{ name }}` (from #1092).

The merged integration case no longer sets `ro: true`. It uses `group` + `use_ipv4acl`, which NX-OS does persist.

---

## Test updates (so CI stays honest)

- Uncommented `cli.yaml` in `tasks/main.yaml` (`network_cli` tag). Added `tests/cli/.gitkeep` so `find` on that directory does not fail.
- Expected commands match the new `setval` (no dummy extra community lines).
- Merged `after.communities` expects one merged object per name, including the ACL.
- Command asserts flatten multi-line module output, then compare. Newline handling uses a real `"\n"` var so ansible-test does not treat `\n` as two characters.
- `replaced` / `overridden` ignore `snmp-server user …` commands when comparing. Default SNMP users differ by image (`cisco` / `ansible` / `temp_user`). We still apply the full user config; we just do not pin image-specific add/remove lines. Idempotency is still asserted.
- Unit tests cover: merged idempotency with split group+ACL lines, and parsed merge of group + `ro` + ACL.

---

## What this is not

- **Not** the AAP-87440 delete bug. That is #1092 (`remval`). This PR needs that fix so `replaced`/`overridden` stay green.
- **Not** NX-API enablement. `nxapi.yaml` is still commented.
- **Not** a full community model change. We still key list compare the same way; we only merge facts so have/want are the same shape.

---

## Test plan

- [x] `ansible-test units tests/unit/modules/network/nxos/test_nxos_snmp_server.py` — 24 passed
- [x] `ansible-test sanity` on the changed files — passed
- [x] `ansible-test network-integration --allow-destructive nxos_snmp_server` against lab NX-OS 10.4 (`network_cli`) — **110 ok, 0 failed** (5 ignored on purpose in `empty_config`)
- [ ] CI network_cli job for `nxos_snmp_server`
- [ ] Confirm #1092 is merged (or this PR’s extra remval commits are acceptable) before merge

### DIFF
--- a/changelogs/fragments/fix_snmp_server_community_removal.yaml
+++ b/changelogs/fragments/fix_snmp_server_community_removal.yaml
@@ -1,0 +1,4 @@
+bugfixes:
+  - nxos_snmp_server - Fix community removal on replaced, overridden, and deleted
+    states by adding a remval template so deletes emit a single
+    ``no snmp-server community`` command instead of re-adding the community.
--- a/changelogs/fragments/reenable_snmp_server_integration.yaml
+++ b/changelogs/fragments/reenable_snmp_server_integration.yaml
@@ -1,0 +1,10 @@
+---
+bugfixes:
+  - nxos_snmp_server - Merge multi-line SNMP community configuration into one
+    object per community name so group and ACL attributes round-trip correctly.
+  - nxos_snmp_server - Parse ``ro`` and ``rw`` community access flags from
+    device configuration.
+  - nxos_snmp_server - Emit only configured community attribute lines so empty
+    ``snmp-server community`` commands no longer reset access to read-only.
+minor_changes:
+  - nxos_snmp_server - Re-enable network_cli integration tests.
--- a/docs/cisco.nxos.nxos_snmp_server_module.rst
+++ b/docs/cisco.nxos.nxos_snmp_server_module.rst
@@ -3981,7 +3981,7 @@ Examples
     #   - no snmp-server user snmp_user_2 use-ipv4acl acl3 use-ipv6acl acl4
     #   - no snmp-server host 192.0.2.2 informs version 3 auth NMS
     #   - snmp-server host 192.0.3.2 informs version 3 auth NMS
-    #   - no snmp-server community private group network-admin
+    #   - no snmp-server community private
     #   - snmp-server community secret group network-operator
     #
     # after:
@@ -4164,7 +4164,7 @@ Examples
     #   - no snmp-server host 192.0.2.1 traps version 1 public
     #   - no snmp-server host 192.0.2.1 source-interface Ethernet1/1
     #   - no snmp-server host 192.0.2.2 informs version 3 auth NMS
-    #   - no snmp-server community private group network-admin
+    #   - no snmp-server community private
     #   - no snmp-server community public group network-operator
     #   - no snmp-server enable traps aaa server-state-change
     #   - no snmp-server enable traps system Clock-change-notification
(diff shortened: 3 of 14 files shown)


malikdotexe:
Closing this to avoid a merge deadlock with #1092 (this PR depended on #1092, and #1092 CI cannot fully exercise SNMP until these test fixes land).

The follow-up commits will go onto #1092 instead so we can test everything together.

PR #1101, opened 2026-09-07
nxos_interfaces: add negotiate_auto attribute for auto-negotiation control

jorgenspange:
## Summary

- Adds a new `negotiate_auto` boolean attribute to `nxos_interfaces` so users can manage auto-negotiation on Ethernet interfaces.
- `negotiate_auto: false` configures `no negotiate auto` (disables auto-negotiation).
- `negotiate_auto: true` configures `negotiate auto` (re-enables auto-negotiation).
- Custom handling in `_compare()` is required because the default (enabled) state produces no CLI output — only `no negotiate auto` appears when disabled. The base class `compare()` skips the `bool False vs None` case, so it would never emit `no negotiate auto` without this explicit logic.

## Motivation

It is currently not possible to configure auto-negotiation through the `cisco.nxos.nxos_interfaces` resource module. NX-OS uses `no negotiate auto` to disable it (unlike IOS which uses `speed nonegotiate`), so a new NX-OS-specific attribute is needed.

Example usage:
```yaml
- name: Disable auto-negotiation on Ethernet1/43
  cisco.nxos.nxos_interfaces:
    config:
      - name: Ethernet1/43
        speed: 1000
        duplex: full
        negotiate_auto: false
    state: merged
```

## Changes

- `argspec/interfaces/interfaces.py` — add `negotiate_auto: bool` option
- `rm_templates/interfaces.py` — add parser matching `no negotiate auto` → `negotiate_auto: False`
- `config/interfaces/interfaces.py` — handle `negotiate_auto` in `_compare()` with explicit logic
- `plugins/modules/nxos_interfaces.py` — document the new suboption
- `tests/unit/modules/network/nxos/test_nxos_interfaces.py` — two new unit tests
- `changelogs/fragments/nxos_interfaces_negotiate_auto.yaml` — changelog entry

## Test plan

- [x] `test_nxos_interfaces_negotiate_auto_disable` — verifies `no negotiate auto` is emitted when `negotiate_auto: false` and interface currently has auto-neg enabled (no command in config)
- [x] `test_nxos_interfaces_negotiate_auto_enable` — verifies `negotiate auto` is emitted when `negotiate_auto: true` and interface currently has `no negotiate auto` in config
- [x] All 14 existing unit tests continue to pass
- [x] Tested against a Cisco Nexus 93180YC-FX running NX-OS 10.4(5) — fact gathering, `merged` state (disable and re-enable), and `replaced` state all behave as expected

### DIFF
--- a/changelogs/fragments/nxos_interfaces_negotiate_auto.yaml
+++ b/changelogs/fragments/nxos_interfaces_negotiate_auto.yaml
@@ -1,0 +1,6 @@
+---
+minor_changes:
+  - nxos_interfaces - Add ``negotiate_auto`` boolean attribute to control
+    auto-negotiation on Ethernet interfaces. Set to ``false`` to configure
+    ``no negotiate auto`` and disable auto-negotiation; set to ``true`` to
+    configure ``negotiate auto`` and re-enable it.
--- a/docs/cisco.nxos.nxos_interfaces_module.rst
+++ b/docs/cisco.nxos.nxos_interfaces_module.rst
@@ -278,6 +278,26 @@ Parameters
                     <td class="elbow-placeholder"></td>
                 <td colspan="4">
                     <div class="ansibleOptionAnchor" id="parameter-"></div>
+                    <b>negotiate_auto</b>
+                    <a class="ansibleOptionLink" href="#parameter-" title="Permalink to this option"></a>
+                    <div style="font-size: small">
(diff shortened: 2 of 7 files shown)


mhuin:
recheck
Don't mind me, just testing queue-ing times on zuul

mhuin:
recheck

PR #1102, opened 2026-09-09
fix: nxos_interfaces reports default-shutdown SVIs as enabled

### LINKED ISSUE #83: nxos_interfaces doesn't detect virtual interfaces or virtual interface state
##### SUMMARY

Changing the administrative status of a virtual switch interface (SVI or L3 interface) for interfaces in shutdown mode (enabled: no) to enabled: yes doesn't work. 
Changing the configuration for these has no effect:
 ``` 
  system default switchport shutdown
  system default interface-vlan autostate
```

```
show int vl42
   Vlan667 is down (Administratively down), line protocol is down, autostate disabled

show running-config interface Vlan42 all
  interface Vlan42
     no description
     shutdown
```

however without the 'all' the SVI won't show the shutdown whereas regular interfaces will:

```
show running-config | section ^interface
interface Vlan42
   ip address 1.2.3.1/24
interface Ethernet1/1
  shutdown
```

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

nxos_interfaces
ansible/module_utils/network/nxos/facts/interfaces/interfaces.py

'all' is missing from the 2nd show run command for section interface
(included here to show the fix):
```diff
             data = connection.get("show running-config all | incl 'system default switchport'")
-            data += connection.get('show running-config | section ^interface')
+            data += connection.get('show running-config all | section ^interface')

```

##### ANSIBLE VERSION

```paste below
ansible 2.9.9
```

##### CONFIGURATION

```paste below
(no changes)
```

##### OS / ENVIRONMENT

CentOS 8
Cisco Nexus9000 C9300v Chassis
NXOS: version 9.3(3)

##### STEPS TO REPRODUCE

```yaml
- nxos_interfaces:
    config:
      - name: "Vlan42"
        enabled: yes
```

##### EXPECTED RESULTS

'no shutdown' command issues for interfaces that are Administratively down.

##### ACTUAL RESULTS

Interface stays in shutdown, no command issued.

jorgenspange:
## Summary

Fixes #83

NX-OS never writes `shutdown` to `show running-config` for interfaces in their default shutdown state. Only `no shutdown` appears, for interfaces explicitly brought up. This causes `nxos_interfaces` to always report default-shutdown SVIs as `enabled: true` and attempt to enable them every run.

## Red Hat solution 7137028

Red Hat [solution 7137028](https://access.redhat.com/solutions/7137028) describes the same class of problem for physical switchports when `system default switchport shutdown` is set, and suggests `no system default switchport shutdown` as a prerequisite.

That workaround does not work. Running-config still shows no admin state for the interface after disabling it:

```
! before
interface Ethernet1/17

! after: no system default switchport shutdown
interface Ethernet1/17
```

Nothing changed. NX-OS does not write the default state regardless of what that default is. This is different from IOS, where `shutdown` is always written explicitly.

The workaround also has no effect on SVIs, which have their own default-shutdown behaviour independent of that global setting.

## Fix

Inject `  shutdown` into any `interface VlanXXX` block in the gathered running-config that has neither `shutdown` nor `no shutdown`, before passing it to the parser. The existing rm_templates already handles `shutdown` -> `enabled: false`.

```python
data = re.sub(
    r"(?m)^(interface Vlan\S+\n)((?:[ \t]+.*\n?)*)",
    lambda m: m.group(1) + (m.group(2) if "shutdown" in m.group(2) else "  shutdown\n" + m.group(2)),
    data,
)
```

## Test

Verified on NX-OS (N9K-C93180YC-FX3H). `Vlan1` in default shutdown state is now reported as `enabled: false` and subsequent runs are idempotent.

### DIFF
--- a/plugins/module_utils/network/nxos/facts/interfaces/interfaces.py
+++ b/plugins/module_utils/network/nxos/facts/interfaces/interfaces.py
@@ -57,7 +57,8 @@ class InterfacesFacts(object):
 
         data = re.sub(
             r"(?m)^(interface Vlan\S+\n)((?:[ \t]+.*\n?)*)",
-            lambda m: m.group(1) + (m.group(2) if "shutdown" in m.group(2) else "  shutdown\n" + m.group(2)),
+            lambda m: m.group(1)
+            + (m.group(2) if "shutdown" in m.group(2) else "  shutdown\n" + m.group(2)),
             data,
         )
 
--- a/plugins/module_utils/network/nxos/facts/interfaces/interfaces.py
+++ b/plugins/module_utils/network/nxos/facts/interfaces/interfaces.py
@@ -15,6 +15,8 @@ for a given resource, parsed, and the facts tree is populated
 based on the configuration.
 """
 
+import re
+
 from ansible_collections.ansible.netcommon.plugins.module_utils.network.common import (
     utils,
 )
@@ -52,6 +54,12 @@ class InterfacesFacts(object):
 
         if not data:
             data = self._get_interface_config(connection)
+
+        data = re.sub(
+            r"(?m)^(interface Vlan\S+\n)((?:[ \t]+.*\n?)*)",
+            lambda m: m.group(1) + (m.group(2) if "shutdown" in m.group(2) else "  shutdown\n" + m.group(2)),
+            data,
+        )
 
         # parse native config using the Interfaces template
         interfaces_parser = InterfacesTemplate(lines=data.splitlines(), module=self._module)


TheNova22:
Hey @jorgenspange! Thanks for this, here is some inputs on the same.
1. A changelog fragment is needed for any change being raised
2. The regex can be a lot more stricter, so something like this maybe: `lambda m: m.group(1) + (m.group(2) if re.search(r"(?m)^[ \t]+(?:no[ \t]+)?shutdown(?:[ \t]*$)", m.group(2)) else "  shutdown\n" + m.group(2))`
3. Regression coverage for explicit and implicit SVI state

PR #1105, opened 2026-09-09
Rewrite nxos_vpc as a modern resource module

jorgenspange:
## Description

Rewrites `nxos_vpc` as a modern resource module using the `rm_base` pattern, consistent with `nxos_ospf_interfaces`, `nxos_vrf_interfaces`, and `nxos_hostname`. Adds all 7 states: `merged`, `replaced`, `overridden`, `deleted`, `gathered`, `rendered`, and `parsed`.

NX-OS-specific handling included:
- `terminal dont-ask` injected before `peer-gateway` / `peer-switch` commands
- composite `peer-keepalive` command with optional `source` and `vrf`
- `auto-recovery` boolean with `no` form

## Type of Change

Refactoring (drop-in replacement, no breaking argspec changes)

## Component Name

`nxos_vpc`

## Testing

Tested against NX-OS 10.5(5) on N9K-C93180YC-FX3H. All states exercised against live device config and working.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ Name | Description
 [cisco.nxos.nxos_udld_interface](https://github.com/ansible-collections/cisco.nxos/blob/main/docs/cisco.nxos.nxos_udld_interface_module.rst)|Manages UDLD interface configuration params.
 [cisco.nxos.nxos_user](https://github.com/ansible-collections/cisco.nxos/blob/main/docs/cisco.nxos.nxos_user_module.rst)|Manage the collection of local users on Nexus devices
 [cisco.nxos.nxos_vlans](https://github.com/ansible-collections/cisco.nxos/blob/main/docs/cisco.nxos.nxos_vlans_module.rst)|VLANs resource module
-[cisco.nxos.nxos_vpc](https://github.com/ansible-collections/cisco.nxos/blob/main/docs/cisco.nxos.nxos_vpc_module.rst)|Manages global VPC configuration
+[cisco.nxos.nxos_vpc](https://github.com/ansible-collections/cisco.nxos/blob/main/docs/cisco.nxos.nxos_vpc_module.rst)|VPC global domain resource module.
 [cisco.nxos.nxos_vpc_interface](https://github.com/ansible-collections/cisco.nxos/blob/main/docs/cisco.nxos.nxos_vpc_interface_module.rst)|Manages interface VPC configuration
 [cisco.nxos.nxos_vrf](https://github.com/ansible-collections/cisco.nxos/blob/main/docs/cisco.nxos.nxos_vrf_module.rst)|(deprecated, removed after 2026-07-25) Manages global VRF configuration.
 [cisco.nxos.nxos_vrf_address_family](https://github.com/ansible-collections/cisco.nxos/blob/main/docs/cisco.nxos.nxos_vrf_address_family_module.rst)|Resource module to configure VRF address family definitions.
--- a/changelogs/fragments/nxos_vpc_resource_module.yaml
+++ b/changelogs/fragments/nxos_vpc_resource_module.yaml
@@ -1,0 +1,11 @@
+breaking_changes:
+  - >-
+    nxos_vpc - rewrite as a modern resource module using the ``rm_base``
+    pattern with all 7 states (``merged``, ``replaced``, ``overridden``,
+    ``deleted``, ``gathered``, ``rendered``, ``parsed``). The flat top-level
+    parameters (``domain``, ``role_priority``, ``system_priority``,
+    ``pkl_dest``, ``pkl_src``, ``pkl_vrf``, ``peer_gw``, ``peer_sw``,
+    ``auto_recovery``, ``delay_restore``) are replaced by a ``config`` dict.
+    Update playbooks from ``domain: 100`` to ``config: {domain: "100", ...}``.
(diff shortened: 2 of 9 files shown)


malikdotexe (changes requested):
Thanks for taking this on, moving `nxos_vpc` onto `rm_base` is the right direction and matches how `nxos_hostname` / `nxos_ospf_interfaces` are already structured. 

### What's working well

- Skeleton is right: argspec / config / facts / rm_templates plus the `facts.py` registration.
- `rendered` and `parsed` work as expected.
- `gathered` works once a domain already exists.
- `merged` of `peer_sw` produced correct CLI, including `terminal dont-ask` where needed.

I went through the diff against `main`, compared it with `hostname` and `ospf_interfaces`, and ran live playbooks against a lab NX-OS **10.4(2)** box to see how it behaves outside of a code read.

### Things to fix before merge

**Empty-device / facts crash**
`domain` is required under `config`, but facts always calls `validate_config` even when parse comes back as `{}`. This means `merged` / `gathered` / `deleted` fail on a device with no VPC domain yet, on the lab this surfaced as a fairly opaque action-plugin error rather than a clean message. Might be worth skipping validation when there's no domain and returning `facts["vpc"] = {}` in that case.

**`replaced` sends an invalid command on 10.4(2)**
Hit `% Incomplete command` on `no system-priority`. Worth double checking the no-forms against the platforms you're targeting, legacy handled this with a default keymap (`system-priority 32667`, `delay-restore 60`) rather than a bare no-form.

**Peer-keepalive still merges under `replaced`**
`_compare` merges have into want whenever any pkl key is present. The docs example for `replaced` shows omitted keys like `pkl_vrf` dropping off, but that's not quite what happens in practice. Might be cleaner to merge only for `merged`/`rendered`, and go want-only for `replaced`/`overridden`.

**Idempotency with `pkl_vrf: management`**
`gathered` doesn't pick this up since NX-OS doesn't nvgen the default VRF, so a follow-up `merged` with the same intent re-pushes the keepalive line every time.

### Tests

`test_nxos_vpc.py` predates this rewrite and patches `get_config` / `load_config`, which no longer exist on the module, so it can't run as-is. Needs replacing with hostname-style unit tests covering all 7 states, plus empty-have, pkl merge vs replace, and boolean `false` when the line is absent from running-config.

The integration target also still calls `state: present` with the old flat `domain:` argspec, so `tests/integration/targets/nxos_vpc` will need rewriting to `config:` + the 7 states, with `feature vpc` enabled in setup and commands/gathered asserted after each change.

### Current CI status

- ✅ Sanity, lint, changelog, and build/import are all green on the stable matrix.
- Unit tests fail across the board, this is the legacy suite erroring in `setUp` (it references methods that no longer exist), rather than new tests catching a regression. More a signal that this needs fresh unit tests than an active bug to chase.
-  Integration ran 12 targets but didn't reach `nxos_vpc`, the suite runs alphabetically and ran out of runway before getting there, so there's no integration coverage from CI yet, hence the lab testing above.

### One more thing

The changelog fragment correctly marks this as breaking, but the PR description still reads like a drop-in/non-breaking change. Would be good to align those so nobody's misled skimming just the description.

Nice work overall, this is a solid base to build from. Happy to take another look once the above lands.

jorgenspange:
Thanks for the review. Pushed 30d0cdd0 addressing all of it, and I rewrote the PR description because this is a breaking change and the original description undersold that.

**Facts crash on a device with no VPC domain** — `validate_config` was being handed an empty dict while `domain` is a required suboption. Facts now short-circuit to `{}` when no domain is present, so `gathered` and `parsed` work on a clean device.

**`no` forms NX-OS rejects** — the numeric sub-commands take a required argument, so `no system-priority` and friends come back as `% Incomplete command`. Removal is now expressed by re-applying the platform default (the old `PARAM_TO_DEFAULT_KEYMAP` values: 32667, 60, 10, 0, 240). Same reasoning applied to `peer-keepalive`, where the negation is now the full line rather than a bare `no peer-keepalive`.

**peer-keepalive merge vs replace** — under `merged` and `rendered` the want layers onto whatever is on the device, so unspecified source or VRF is retained. Under `replaced` and `overridden` the want stands alone. An absent VRF is normalised to `management` on both sides, so `pkl_vrf: management` is idempotent even though NX-OS does not nvgen it.

**Boolean false** — an absent line now reads as `false` on both sides, so `peer_gw: false` on a device without the line is idempotent, while the `no` form is still emitted when the line is present.

**Domain ID change** — `replaced` and `overridden` now tear down the old domain first with `terminal dont-ask` and `no vpc domain <old>` before configuring the new one.

**Unit tests** — replaced. 20 tests covering all 7 states plus the default-reset, peer-keepalive layering, VRF normalisation, and boolean-false cases. Full nxos suite passes.

**Integration target** — rewritten from `sanity.yaml` to `config:` plus the 7 states in the `nxos_hostname` layout. `ansible-lint --profile production` passes.

**Breaking changes recorded in the changelog fragment** — `state: present` / `state: absent` are gone in favour of `merged` / `deleted`, and passing the literal `default` as a numeric value is no longer supported (omit the attribute under `replaced` or `overridden` instead).

On the coordination note: #1106 is now independent of this one. Its integration target builds the VPC domain with `nxos_config` rather than calling `nxos_vpc`, so the two can land in either order. Whichever merges second I will rebase and confirm both `vpc=` and `vpc_interfaces=` survive in `facts.py`.

PR #1106, opened 2026-09-09
Rewrite nxos_vpc_interface as a modern resource module

jorgenspange:
## Description

Rewrites `nxos_vpc_interface` as a modern resource module using the `rm_base` pattern, consistent with `nxos_ospf_interfaces` and `nxos_vrf_interfaces`. Adds all 7 states: `merged`, `replaced`, `overridden`, `deleted`, `gathered`, `rendered`, and `parsed`.

Changes from the legacy module:
- Uses `name: port-channel10` (full interface name) as the identifier, matching the convention of other interface modules
- Adds `orphan_port_suspend` support (`vpc orphan-port suspend`), which works on any interface type including Ethernet
- Facts gathered via `show running-config | section ^interface`; only interfaces with VPC-related config are returned

## Type of Change

Refactoring (drop-in replacement with minor argspec alignment to collection conventions)

## Component Name

`nxos_vpc_interface`

## Testing

Tested against NX-OS 10.5(5) on N9K-C93180YC-FX3H. All states exercised against live device config and working.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ Name | Description
 [cisco.nxos.nxos_user](https://github.com/ansible-collections/cisco.nxos/blob/main/docs/cisco.nxos.nxos_user_module.rst)|Manage the collection of local users on Nexus devices
 [cisco.nxos.nxos_vlans](https://github.com/ansible-collections/cisco.nxos/blob/main/docs/cisco.nxos.nxos_vlans_module.rst)|VLANs resource module
 [cisco.nxos.nxos_vpc](https://github.com/ansible-collections/cisco.nxos/blob/main/docs/cisco.nxos.nxos_vpc_module.rst)|Manages global VPC configuration
-[cisco.nxos.nxos_vpc_interface](https://github.com/ansible-collections/cisco.nxos/blob/main/docs/cisco.nxos.nxos_vpc_interface_module.rst)|Manages interface VPC configuration
+[cisco.nxos.nxos_vpc_interface](https://github.com/ansible-collections/cisco.nxos/blob/main/docs/cisco.nxos.nxos_vpc_interface_module.rst)|VPC interface resource module
 [cisco.nxos.nxos_vrf](https://github.com/ansible-collections/cisco.nxos/blob/main/docs/cisco.nxos.nxos_vrf_module.rst)|(deprecated, removed after 2026-07-25) Manages global VRF configuration.
 [cisco.nxos.nxos_vrf_address_family](https://github.com/ansible-collections/cisco.nxos/blob/main/docs/cisco.nxos.nxos_vrf_address_family_module.rst)|Resource module to configure VRF address family definitions.
 [cisco.nxos.nxos_vrf_af](https://github.com/ansible-collections/cisco.nxos/blob/main/docs/cisco.nxos.nxos_vrf_af_module.rst)|Manages VRF AF.
--- a/changelogs/fragments/nxos_vpc_interface_resource_module.yaml
+++ b/changelogs/fragments/nxos_vpc_interface_resource_module.yaml
@@ -1,0 +1,11 @@
+breaking_changes:
+  - >-
+    nxos_vpc_interface - rewrite as a modern resource module using the
+    ``rm_base`` pattern with all 7 states (``merged``, ``replaced``,
+    ``overridden``, ``deleted``, ``gathered``, ``rendered``, ``parsed``).
+    The flat ``portchannel`` / ``vpc`` / ``peer_link`` parameters are replaced
+    by a ``config`` list where each entry uses ``name: port-channel10``
+    (full interface name) as the identifier. Adds ``orphan_port_suspend``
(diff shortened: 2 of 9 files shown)


jorgenspange:
@NilashishC would you be able to have a look at this one?

malikdotexe (changes requested):
Thanks for this one as well. The list-RM shape is right, keying by `name`, handling deleted/overridden extras, and the parsers cleanly distinguish `vpc N` from `vpc peer-link` and `vpc orphan-port suspend`. Requesting changes for now, mostly around a few gaps rather than anything structural.

### What's working well

- `merged` of a VPC ID onto a port-channel worked end to end on the lab, and was idempotent on a repeat run, the core happy path is solid.
- `rendered` / `parsed` both passed.
- `peer-link` didn't come through on this particular lab (`% Invalid command`), but that's plausibly a topology issue (single node, port-channel with no members) rather than a code issue, flagging it as something to confirm against a real VPC pair rather than a confirmed bug.

### Things to fix before merge

**`merged` can't turn things off**
`peer_link: false` and `orphan_port_suspend: false` both come back as no-ops with empty commands. The old integration target expected `peer_link: false` to actually remove the peer-link. Might be worth following the `ospf_interfaces` pattern, `dict_merge(have, want)` into a single compare path, instead of the current one-directional `elif` handling.

**Safety checks from the legacy module**
A few checks aren't carried over: failing if the port-channel doesn't exist, if a VPC ID is already assigned elsewhere, or the `vpc`/`peer_link` mutual exclusivity (still documented in the argspec but not enforced). Letting the device be the source of truth here is a reasonable design choice for a resource module, just flagging that the current `elif` means `peer_link` can get silently dropped when `vpc` is also set, which might be worth surfacing more clearly instead of swallowing.

### Tests

Same pattern as the `nxos_vpc` PR, `test_nxos_vpc_interface.py` patches `load_config`, which no longer exists on the module, so it can't run as-is once CI gets to it. Needs replacing with hostname-style unit tests covering all 7 states, plus `peer_link: false`, the `vpc`/`peer_link` exclusivity case, and overridden stripping other POs.

The integration target also still uses `portchannel: 10` and calls the legacy `nxos_vpc` API (`state: present`, `domain: 100`), so `tests/integration/targets/nxos_vpc_interface` will need rewriting to `config:` + the 7 states too, similar to what we're asking for on the `nxos_vpc` PR. Worth noting this depends on the `nxos_vpc` rewrite landing first since the interface tests need a domain to exist.

### Current CI status

- ⚠️ Integration ran 12 targets but the alphabetical ordering meant it didn't reach `nxos_vpc_interface` before the run ended, same as the `nxos_vpc` PR. So the lab testing above is what we have to go on for now.

### Coordination note

This and the `nxos_vpc` PR will both try to register a fact class in `facts.py`, whichever lands second should double check both `vpc=` and `vpc_interfaces=` survive the merge.

Good progress, the core path clearly works. Once merged-false handling and unit/integration tests are in (covering `peer_link: false`, the exclusivity case, and overridden stripping other POs), this should be close.

jorgenspange:
Thanks for the thorough review. Pushed 38ef9c91 addressing everything.

**`merged` can't turn things off** — fixed by taking the `ospf_interfaces` approach you suggested. `want` is layered onto `have` with `dict_merge` and there is now one compare path for every state, so `peer_link: false` and `orphan_port_suspend: false` emit `no vpc peer-link` / `no vpc orphan-port suspend`. Attributes the playbook omits are still inherited from the device, so `merged` keeps its usual meaning.

**`vpc` / `peer_link` exclusivity** — now enforced in the argspec with a suboption-level `mutually_exclusive`, so supplying both in one `config` entry fails the task rather than quietly dropping `peer_link`. Related: asking for one of the pair explicitly clears the other on the device, so switching a port-channel from `vpc 10` to the peer-link emits `no vpc` before `vpc peer-link` instead of pushing both. Documented in the module notes and the changelog fragment.

**Other legacy safety checks** — left out deliberately, going with device-as-source-of-truth as you suggested. The PR description now says so explicitly so it is a recorded decision rather than an omission.

**Unit tests** — replaced. 22 tests covering all 7 states, `peer_link: false`, `orphan_port_suspend: false`, both directions of the `vpc`/`peer-link` switch-over, the exclusivity failure, and `overridden` stripping other port-channels. Full nxos suite passes at 1253 tests.

**Integration target** — rewritten to `config:` plus the 7 states in the `nxos_hostname` layout. Setup builds the VPC domain and the test port-channels with `nxos_config` rather than calling `nxos_vpc`, so this target no longer depends on which `nxos_vpc` API is in the tree and the two PRs can land in either order. `peer_link.yaml` covers configuring the peer-link and removing it via `merged`. The peer-link port-channel is created as a trunk, which may be what the `% Invalid command` in your lab was about. `ansible-lint --profile production` passes.

**Coordination note** — on this branch `facts.py` registers `vpc_interfaces=` only, since the `vpc` facts class lives on #1105. Whichever merges second, I will rebase and confirm both registrations survive.

